PR #22, opened 2021-07-28
Track well known principals and domain controllers and create appropriate output objects for them.

rvazarkar:


src/CommonLib/Extensions.cs
```diff
@@ -244,6 +244,9 @@ public static Label GetLabel(this SearchResultEntry entry)
 
             var objectId = entry.GetObjectIdentifier();
 
+            if (objectId == null)
```
ddlees, line 247: So weird that oid would be null but good catch

src/CommonLib/LDAPUtils.cs
```diff
@@ -84,6 +132,17 @@ public string ConvertWellKnownPrincipal(string sid, string domain)
             return $"{forest}-{sid}".ToUpper();
         }
 
+        private async Task<Group> GetBaseEnterpriseDC()
+        {
+            var forest = GetForest()?.Name;
+            if (forest == null) Logging.Debug("Error getting forest, ENTDC sid is likely incorrect");
```
ddlees, line 138: Since the message is advisory we log this at the warn level

src/CommonLib/LDAPUtils.cs
```diff
@@ -84,6 +132,17 @@ public string ConvertWellKnownPrincipal(string sid, string domain)
...
+            var g = new Group {ObjectIdentifier = $"{forest}-S-1-5-9".ToUpper()};
```
ddlees, line 139: In the case where `forest` is `null` the OID will be `-S-1-5-9`. Is that what we want or do we want to deliberately prefix with some string?

PR #24, opened 2021-10-12
Remove newtonsoft

rvazarkar:
Remove the dependency on newtonsoft.json

### DIFF
--- a/src/CommonLib/Cache.cs
+++ b/src/CommonLib/Cache.cs
@@ -150,37 +150,6 @@ namespace SharpHoundCommonLib
             return CacheInstance;
         }
 
-        // public static void LoadExistingCache(string filePath)
-        // {
-        //     if (!File.Exists(filePath))
-        //     {
-        //         CacheInstance = new Cache();
-        //         Logging.Debug("Cache file not found, empty cache created.");
-        //         return;
-        //     }
-        //
-        //     try
-        //     {
-        //         Logging.Debug($"Loading cache from {filePath}");
-        //         var bytes = File.ReadAllBytes(filePath);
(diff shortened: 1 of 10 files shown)


src/CommonLib/Cache.cs
```diff
@@ -125,54 +122,64 @@ private static string GetPrefixKey(string key, string domain)
...
-
-            Logging.Debug(
-                $"Cache file loaded!\n {GetCacheStats()}");
-        }
+        public static Cache GetCacheInstance()
+        {
+            return CacheInstance;
+        }
+
+        // public static void LoadExistingCache(string filePath)
```
ddlees, line 153: remove commented out code or include a comment

src/CommonLib/Cache.cs
```diff
@@ -184,16 +191,16 @@ private static void CreateMissingDictionaries()
...
-            var serialized = new UTF8Encoding(true).GetBytes(JsonConvert.SerializeObject(CacheInstance,
-                new JsonSerializerSettings
-                {
-                    DefaultValueHandling = DefaultValueHandling.Include
-                }));
-            using var stream = new FileStream(filePath, FileMode.Create, FileAccess.Write, FileShare.None);
-            stream.Write(serialized, 0, serialized.Length);
-            Logging.Log(LogLevel.Information, $"Wrote cache file to {filePath}\n{GetCacheStats()}");
-        }
+        // public static void SaveCache(string filePath)
```
ddlees, line 194: remove commented out code or leave a comment

src/CommonLib/Cache.cs
rvazarkar: Luckily, no one uses this yet. I will be bumping the version to 1.1.0 though

PR #26, opened 2022-02-18
fix: ensure highlevel is being set on all objects

ddlees:
Addresses https://github.com/BloodHoundAD/BloodHound/issues/505


ddlees:
After some consideration it's probably best to implement this in https://github.com/BloodHoundAD/SharpHound to properly hit all objects

PR #28, opened 2022-04-26
Minor Change to Computer ACL

api0cradle:
Added AllExtendedRights to computers even if LAPS is not installed in the environment.

### DIFF
--- a/src/CommonLib/Processors/ACLProcessor.cs
+++ b/src/CommonLib/Processors/ACLProcessor.cs
@@ -253,6 +253,7 @@ namespace SharpHoundCommonLib.Processors
                         RightName = EdgeNames.AddSelf
                     };
 
+
                 //Process object type specific ACEs. Extended rights apply to users, domains, and computers
                 if (aceRights.HasFlag(ActiveDirectoryRights.ExtendedRight))
                 {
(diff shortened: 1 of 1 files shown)


api0cradle:
Blog post https://www.trustedsec.com/blog/diving-into-pre-created-computer-accounts/

yellow-starburst:
@ddlees

andyrobbins:
@api0cradle A few questions:

1. Is the AllExtendedRights ACE only abusable if "Assign this computer account as a pre-Windows 2000 computer" is set to true?
2. When you use NetUserChangePassword or Kpasswd to reset the computer account password, does that break the trust between the computer and AD, or does the password change trickle down to the computer as well?
3. Does AllExtendedRights against a computer also allow you to perform RBCD? 

Andy

api0cradle:
`1. Is the AllExtendedRights ACE only abusable if "Assign this computer account as a pre-Windows 2000 computer" is set to true?`
The group/user you choose when creating a computer account will have the AllExtendedRights regardless of the "Assign this computer account as pre-Windows 2000 computer". 

`2. When you use NetUserChangePassword or Kpasswd to reset the computer account password, does that break the trust between the computer and AD, or does the password change trickle down to the computer as well?`
This breaks the trust. Password is not replicated down so abusing the AllExtendedRights (reset/change password) would break the trust between the computer and the domain.
In my blog post I am targeting unused pre-created computer accounts and of course there is a responsibility on the tester to verify (as with all attacks) that it is okay to actually perform the attack after AllExtendedRights is identified. 

`3. Does AllExtendedRights against a computer also allow you to perform RBCD?`
This would be same scenario as before (same flow as documented attacks), except that you are not creating a computer account to perform the attack, instead taking over an existing. So if someone finds a computer account that they have AllExtendedRights they could change the password and use it to perform a RBCD attack.
However, it would of course be up to the operator to figure out if that is really a path worth taking since you can break the trust relationship (as mentioned in 2). 



The AllExtendedRights are already gathered today by SharpHound if the environment has LAPS installed, so in my opinion this is not a big change in functionality if this PR would be approved.

Acebond:
Any reason this was closed? I wanna know about AllExtendedRights on computer objects even when LAPS doesn't exist.

JonasBK:
It closed when @api0cradle deleted his fork of SharpHoundCommon, as the PR came from that fork.

PR #30, opened 2022-06-18
Add an option to disable LDAPS Certificate verification

Augustin-FL:
By default, DotNet (which SharpHound uses) perform strong verification of LDAPS TLS certificates. 
This is unlike `bloodhound-python` which [does not verify SSL](https://github.com/cannatag/ldap3/blob/master/ldap3/core/tls.py#L73) on its queries. 

DotNet TLS verification [is notoriously tricky](https://stackoverflow.com/questions/58165262/net-program-does-not-get-validated-server-certificate), and [sometimes fail](https://stackoverflow.com/questions/28761249/x509certificate2-verify-returns-false-always) even trough the certificate is perfectly valid (For instance, because the CRL cannot be reached or because TLS 1.3 can't be used..). This can be a problem when running bloodhound in LDAPS-only environments.

This PR add an option to disable verification of the TLS certificate when doing LDAPS queries

(Related to https://github.com/BloodHoundAD/SharpHound/pull/24)

### DIFF
--- a/src/CommonLib/LDAPConfig.cs
+++ b/src/CommonLib/LDAPConfig.cs
@@ -10,6 +10,7 @@ namespace SharpHoundCommonLib
         public int Port { get; set; } = 0;
         public bool SSL { get; set; } = false;
         public bool DisableSigning { get; set; } = false;
+        public bool DisableCertVerification { get; set; } = false;
         public AuthType AuthType { get; set; } = AuthType.Kerberos;
 
         public int GetPort()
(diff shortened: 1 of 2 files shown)


ddlees:
recheck

ddlees:
@Augustin-FL In an attempt to accept contributions the "right" way and be the best stewards we can be to the FOSS community we have recently published a Contributor License Agreement. Would you please consider reading and signing the CLA linked in the comment above? Thank you!

Augustin-FL:

PR #32, opened 2022-07-04
Update PortScanner.cs

yellow-starburst:
change default value from 1/2 second to 2 seconds.
Fixes https://github.com/BloodHoundAD/BloodHound/issues/548


rvazarkar:
We're going to reject this in favor of just using the flag thats available.

PR #33, opened 2022-07-11
Add the ForceChangePassword attack path for Computer node

CravateRouge:
The AllExtendedRights and UserForceChangePassword rights allow also to change the password of the computer. It will probably not be seen in production environment but can be useful in CTF.

### DIFF
--- a/src/CommonLib/Processors/ACLProcessor.cs
+++ b/src/CommonLib/Processors/ACLProcessor.cs
@@ -304,26 +304,31 @@ namespace SharpHoundCommonLib.Processors
                     }
                     else if (objectType == Label.Computer)
                     {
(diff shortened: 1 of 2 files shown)


CravateRouge:
@rvazarkar I think it's useful to see this edge case event if it's uncommon. I want to be able to cover it if it happens during my tests

rvazarkar:
So I think we do want to support this, but we're going to want to add this as a discrete edge called ResetComputerPassword and make it abundantly clear that this could likely cause issues if used in real environments unless handled very carefully

ZkClown:
Hello @rvazarkar ,
During an assessment, my team and I have found this edge on a test computer that the enterprise had created and almost missed it if we had not checked the ACE using other tools (like bloodyAD). I think it could be cool to have this implemented. I know that BloodHound.py which is the python ingestor handles it.
Best regards.

PR #34, opened 2022-07-11
Add AddAllowedToAct logic based on User Account Restrictions property set

dirkjanm:
This property set contains the `msDS-AllowedToActOnBehalfOfOtherIdentity` property so if you have write privs on this set you can configure RBCD.

Will add a link to a blog soon with more context.

### DIFF
--- a/src/CommonLib/EdgeNames.cs
+++ b/src/CommonLib/EdgeNames.cs
@@ -19,5 +19,6 @@
         public const string WriteSPN = "WriteSPN";
         public const string AddKeyCredentialLink = "AddKeyCredentialLink";
         public const string SQLAdmin = "SQLAdmin";
+        public const string WriteAccountRestrictions = "WriteAccountRestrictions";
     }
 }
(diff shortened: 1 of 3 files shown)


dirkjanm:
Blog with context: https://dirkjanm.io/abusing-forgotten-permissions-on-precreated-computer-objects-in-active-directory/

rvazarkar:
So after discussing this internally, this is some really cool stuff, but we'd like to extract this into a new edge.  This is mainly because its hard to remediate without additional details that are related to what's actually happening on the backend. Feel free to propose a new edge name, but we love the work.

Microsoft docs being wrong again is always fun

rvazarkar:
Thanks!

ddlees:
recheck

ddlees:
@dirkjanm In an attempt to accept contributions the "right" way and be the best stewards we can be to the FOSS community we have recently published a Contributor License Agreement. Would you please consider reading and signing the CLA linked in the comment above? Thank you!

PR #39, opened 2022-09-20
Support exceptions in QueryLDAP methods

superlinkx:
This pull request adds support for throwing exceptions in all `QueryLDAP()` methods. Because this is only a "sometimes applicable" feature for tracing exact problems, it adds an optional flag at the end of the method calls called `throwException`. The default behavior remains the same and will log errors exactly as before. However, when `throwException` is enabled, any code that pumps the iterator will need to catch exceptions.

Exceptions are of a new type `SharpHoundCommonException` and will contain the details that would typically be logged in normal operation. This will allow for calling code to examine exceptions and then make informed decisions on how to proceed. In the default configuration, a warning will be logged on any failure, but no additional results will be returned.

This pull request includes a basic test to ensure that both states of the flag work correctly (throwing an exception or not) and that they work with the "options" variant of `QueryLDAP`.

### DIFF
--- a/src/CommonLib/Exceptions/SharpHoundCommonException.cs
+++ b/src/CommonLib/Exceptions/SharpHoundCommonException.cs
@@ -1,10 +1,11 @@
 using System;
 
-namespace SharpHoundCommonLib.Exceptions;
-
-public class SharpHoundCommonException : Exception
+namespace SharpHoundCommonLib.Exceptions
 {
-    public SharpHoundCommonException() { }
-    public SharpHoundCommonException(string message) : base(message) { }
-    public SharpHoundCommonException(string message, Exception inner) : base(message, inner) { }
+    public class SharpHoundCommonException : Exception
+    {
+        public SharpHoundCommonException() { }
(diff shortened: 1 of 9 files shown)


superlinkx:

src/CommonLib/Exceptions/SharpHoundCommonException.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace SharpHoundCommonLib.Exceptions
+{
+    public class SharpHoundCommonException : Exception
```
rvazarkar, line 5: Can we rename this to something thats more specific to our problem set, like maybe LDAPQueryException or something. SharpHoundCommonException is a bit too generic

superlinkx: Absolutely. I think LDAPQueryException probably makes the most sense here honestly, so I'll refactor that quick.

superlinkx:
Closing in favor of #40

PR #45, opened 2022-12-08
Fix for null sid values when resolving hosts to a sid

mwstock:
SharpHound does not populate session information due to a null value being assigned as the SID in some cases. This appears to be due to one step in the ResolveHostToSid function where the ObjectIdentifier from the returned TypedPrincipal object is not assigned to the sid variable before being stored in the _hostResolutionMap dictionary. This fix should match the previous behaviour observed in SharpHound3.


mwstock:

rvazarkar:
Thanks for the MR, will make it into the next version

PR #46, opened 2022-12-20
DumpSMSAPassword: collection and resolving of msds-hostserviceaccount

simondotsh:
Collection of the attribute `msds-hostserviceaccount` on computer objects when reading properties. It is not added to the list of properties that are shown on the UI when viewing an object since it is only collected in order to build edges. The distinguished names contained in the attribute are resolved, and only object identifiers are returned to avoid bloating with unnecessary data.

For more info, see https://github.com/BloodHoundAD/BloodHound/issues/624.


simondotsh:
Hi @rvazarkar,

I added Jonas' unit test and null check to pass the test. Let me know if there's anything else.

PR #47, opened 2023-02-18
Added option to do Session Enumeration as local Admin User

LuemmelSec:
This is the needed PR to SharpHoundCommon that allows to set an option to do session enumeration with a dedicated local (admin) user.
Credits to @eversinc33 how helped me a lot with this.

The situation with more recent pentests and red-team engagements is most likely that you see higher numbers of Windows Server 2016 + as well as Windows 10 1607 +. So the privileged session enumeration via NetWkstaUserEnum needs to be with a user that has local admin rights. Once gaining a foothold or in an assumed breach scenario we will most likely do not have these rights.
But, and that is very common, we do local priv esc on the client, fetch the local admin's credentials, and it happens to be some lame password that is reused accross a large scope of systems - worst case even servers. Same applies to once we get a server's local admin password, and this one is reused.

SharpHound was not able to do this, because it would run the session checks in the context of  the user that runs the collector.
What we did was to add 3 new flags:

-DoLocalAdminSessionEnum
-LocalAdminUsername
-LocalAdminPassword

We added a new library (Impersonate.cs ) that allows us to impersonate another remote local user in the current user's context via the LOGON32_PROVIDER_WINNT50 option from the advapi32.dll. 
That allows us to wrap the session enumeration stuff inside an impersonated user's context and overcome the limitations mentioned above.
Credits to @phillipharding for this library.

Everything else is done as the running user, so no impact on the rest of what SharpHound is doing.

I created a POC video demonsrating how it works:
https://www.youtube.com/watch?v=mMpNs3MXISM

BloodHound documentation was extended accordingly as well as the SharpHound project itself. Accordings PRs were opened.

https://github.com/BloodHoundAD/SharpHound/pull/40
https://github.com/BloodHoundAD/BloodHound/pull/646

### DIFF
--- a/src/CommonLib/Impersonate.cs
+++ b/src/CommonLib/Impersonate.cs
@@ -1,0 +1,198 @@
+﻿using System;
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+using System.Security.Principal;
+using System.Xml.Linq;
+
+namespace Impersonate
+{
+    public enum LogonType
+    {
+        LOGON32_LOGON_INTERACTIVE = 2,
+        LOGON32_LOGON_NETWORK = 3,
+        LOGON32_LOGON_BATCH = 4,
+        LOGON32_LOGON_SERVICE = 5,
+        LOGON32_LOGON_UNLOCK = 7,
+        LOGON32_LOGON_NETWORK_CLEARTEXT = 8, // Win2K or higher
+        LOGON32_LOGON_NEW_CREDENTIALS = 9 // Win2K or higher
+    };
+
+    public enum LogonProvider
+    {
+        LOGON32_PROVIDER_DEFAULT = 0,
+        LOGON32_PROVIDER_WINNT35 = 1,
+        LOGON32_PROVIDER_WINNT40 = 2,
+        LOGON32_PROVIDER_WINNT50 = 3
+    };
+
+    public enum ImpersonationLevel
+    {
+        SecurityAnonymous = 0,
+        SecurityIdentification = 1,
+        SecurityImpersonation = 2,
+        SecurityDelegation = 3
+    }
+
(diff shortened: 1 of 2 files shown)


eversinc33:

eversinc33:
recheck

rvazarkar:
This is an awesome PR, I've gotta dig into this a bit

LuemmelSec:
> This is an awesome PR, I've gotta dig into this a bit

Thanks Rohan.
There were already people asking if local Admin enum would now also be possible:
https://twitter.com/NotMedic/status/1626968323920482306

This ofcourse could now be wrapped too, as the library is there. Accordingly every function that would need it as well.

JonasBK:
Hi @phillipharding,

I noticed you have been given credit for creating the Impersonate.cs file in this PR: https://github.com/BloodHoundAD/SharpHoundCommon/pull/47/files#diff-4121b9ee7af13bcce5901c1278346c5985d7964db39d80f718b0ffb2ff35a81b

Can you confirm that you are ok with your work being used in this project? :)

BR Jonas

LuemmelSec:
> Hi @phillipharding,
> 
> I noticed you have been given credit for creating the Impersonate.cs file in this PR: https://github.com/BloodHoundAD/SharpHoundCommon/pull/47/files#diff-4121b9ee7af13bcce5901c1278346c5985d7964db39d80f718b0ffb2ff35a81b
> 
> Can you confirm that you are ok with your work being used in this project? :)
> 
> BR Jonas

To add to this for the sake of completeness it was found here:
https://stackoverflow.com/questions/125341/how-do-you-do-impersonation-in-net

Which linked to here:
https://web.archive.org/web/20150315063000/https://platinumdogs.me/2008/10/30/net-c-impersonation-with-network-credentials/

PR #50, opened 2023-02-28
LDAP Query Semaphore

rvazarkar:
Add a semaphore to QueryLDAP to limit maximum number of concurrent searches to prevent Server is Busy errors


src/CommonLib/LDAPUtils.cs
```diff
@@ -370,8 +374,16 @@ public IEnumerable<string> DoRangedRetrieval(string distinguishedName, string at
             {
                 SearchResponse response;
 
-                response = (SearchResponse) conn.SendRequest(searchRequest);
-
+                try
+                {
+                    _searchQuerySemaphore.WaitOne();
```
zinic, line 379: I have some minor concerns around waits like this. This could result in a lock up in the daemon that can only be resolved by user intervention (a restart). There are other `WaitOne(...)` methods that support setting a timeout. We may want to consider utilization of one of those while consuming the boolean return value to gate further use of the connection.

This introduces additional error handing concerns but at the benefit of not falling into deadlock.

zinic (changes requested):
Left a comment on the use of `WaitOne(...)`. Otherwise the changes look sane to me.

src/CommonLib/LDAPUtils.cs
zinic: Maybe on boolean `false` throw an exception or something else that will get caught and interpreted at a layer above.

zinic (approved):
I have some mild fear/concern around a semaphore signal not being sent due to resource acquisition being handled in one try-catch while release is in another but I don't have a better solution at this moment that doesn't require boiling the ocean. This looks good, otherwise.

PR #55, opened 2023-04-10
fix: remove LDAP semaphore since it didn't actually fix anything. Sad.

rvazarkar:
fix: catch an exception case in GetLocalDomainInformation


zinic:
:+1:

PR #56, opened 2023-04-19
Allow retries for "Server is Busy" ldap queries.

rvazarkar:


src/CommonLib/LDAPUtils.cs
```diff
@@ -378,13 +472,25 @@ public IEnumerable<string> DoRangedRetrieval(string distinguishedName, string at
...
+            var retryCount = 0;
+            
             while (true)
             {
                 SearchResponse response;
                 try
                 {
                     response = (SearchResponse) conn.SendRequest(searchRequest);
                 }
+                catch (LdapException le) when (le.ErrorCode == 51 && retryCount < MaxRetries)
```
ddlees, line 485: Not a huge deal but it might be worth extracting ldap error codes like this into a constant to make `51` a little more understandbale.

src/CommonLib/LDAPUtils.cs
```diff
@@ -1428,6 +1551,55 @@ internal string ResolveDomainNetbiosToDns(string domainName)
 
             return domainName.ToUpper();
         }
+        
+        /// <summary>
+        /// Gets the range retrieval limit for a domain
+        /// </summary>
+        /// <param name="domainName"></param>
+        /// <returns></returns>
+        public int GetDomainRangeSize(string domainName = null)
```
ddlees, line 1560: Consider changing this to `GetDomainRangeSize(string domainName = null, int defaultSize = 750)`.

ddlees: Could also use an associated test if you've got the bandwidth

superlinkx (approved):

PR #60, opened 2023-04-21
New container enumeration

rvazarkar:
Enumerate object contains using DNs instead of querying container child items directly

### DIFF
--- a/src/CommonLib/Helpers.cs
+++ b/src/CommonLib/Helpers.cs
@@ -24,6 +24,39 @@ namespace SharpHoundCommonLib
             "S-1-5-19", "S-1-5-20", "S-1-0-0", "S-1-0", "S-1-2-1"
         };
 
+        public static string RemoveDistinguishedNamePrefix(string distinguishedName)
+        {
+            if (!distinguishedName.Contains(","))
+            {
+                return "";
+            }
+            if (distinguishedName.IndexOf("DC=", StringComparison.OrdinalIgnoreCase) < 0)
+            {
+                return "";
+            }
+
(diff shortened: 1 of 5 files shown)


src/CommonLib/Helpers.cs
```diff
@@ -24,6 +24,39 @@ public static class Helpers
...
+                //If theres a comma at the beginning of the DN, something screwy is going on. Just ignore it
+                if (i == 0)
+                {
+                    continue;
+                }
+
+                //This indicates an escaped comma
+                if (distinguishedName[i-1] == '\\')
+                {
+                    Console.WriteLine(distinguishedName[i-1]);
```
superlinkx, line 49: This should likely be removed if it was here for debugging or converted to a proper log statement that explains what this entry would be in the actual log

src/CommonLib/Helpers.cs
```diff
@@ -24,6 +24,39 @@ public static class Helpers
...
+            if (!distinguishedName.Contains(","))
+            {
+                return "";
+            }
+            if (distinguishedName.IndexOf("DC=", StringComparison.OrdinalIgnoreCase) < 0)
+            {
+                return "";
+            }
+
+            for (var i = distinguishedName.IndexOf(','); i > -1; i = distinguishedName.IndexOf(',', i + 1))
```
superlinkx, line 38: I found this logic pretty hard to parse. It appears correct after looking through usage and tests but I think I'd want to see a comment at the very least explaining what this for loop is doing and why it's necessary

superlinkx (commented):
I have a couple of small concerns with the helper, but the tests made it clear how it should work at least. It looks good to deploy if we can answer to these two housekeeping comments.

superlinkx (approved):

PR #64, opened 2023-06-12
Logging updates for port scanner

rvazarkar:
Just some quick logging updates to help diagnose issues

### DIFF
--- a/src/CommonLib/Processors/PortScanner.cs
+++ b/src/CommonLib/Processors/PortScanner.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Concurrent;
+﻿using System;
+using System.Collections.Concurrent;
 using System.Net.Sockets;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/CommonLib/Processors/PortScanner.cs
```diff
@@ -37,28 +38,27 @@ public virtual async Task<bool> CheckPort(string hostname, int port = 445, int t
...
             }
 
             try
             {
                 using var client = new TcpClient();
                 var ca = client.ConnectAsync(hostname, port);
-                await Task.WhenAny(ca, Task.Delay(timeout));
-                client.Close();
-                if (!ca.IsFaulted && ca.IsCompleted)
+                if (await Task.WhenAny(ca, Task.Delay(timeout)) == ca)
```
juggernot325, line 49: I just want to call this out as a sanity check. The PR says that this is just logging changes but this appears to be a logic change. Just making sure it was intentional.

Issue #70, opened 2023-09-11
Missing Collection of Group Add/Remove Self As Member


godylockz:
https://book.hacktricks.xyz/windows-hardening/active-directory-methodology/acl-persistence-abuse#self-self-membership-on-group
Self (Self-Membership) - ability to add yourself to a group

In Security Settings:
Permission: Add/remove self as member
Permission: All validated writes

This could be hidden privilege as a "member of a privileged group" and be missed in BloodHound path tracing.

In dsacls.exe, it comes up as:
SPECIAL ACCESS
WRITE SELF

JonasBK:
Hi!

We do have an AddSelf edge which should cover that: https://support.bloodhoundenterprise.io/hc/en-us/articles/17358095502363-AddSelf
Have you experienced that the edge is not created when this permission is granted?

godylockz:
> Hi!
> 
> We do have an AddSelf edge which should cover that: https://support.bloodhoundenterprise.io/hc/en-us/articles/17358095502363-AddSelf Have you experienced that the edge is not created when this permission is granted?

The permission is not captured in any SharpHound collector agent. 
I have tried bloodhound-python, multiple SharpHound versions, and crackmapexec as well.

JonasBK:
@godylockz provided details about the environment in a private chat. Thanks a lot @godylockz!

The AddSelf edge is created when a principal is granted the "Add/remove self as member" privilege. That ACE looks like this:
```
ObjectDN              : CN=T0_Admins,OU=Groups,OU=Tier0,DC=dumpster,DC=fire
InheritedObject       : 
Object                : Member
ActiveDirectoryRights : Self
InheritanceType       : None
ObjectType            : bf9679c0-0de6-11d0-a285-00aa003049e2
InheritedObjectType   : 00000000-0000-0000-0000-000000000000
ObjectFlags           : ObjectAceTypePresent
AccessControlType     : Allow
IdentityReference     : DUMPSTER\addself
IsInherited           : False
InheritanceFlags      : None
PropagationFlags      : None
```

When selecting "All validated writes" in the UI, the "Add/remove self as member" is automatically selected as well. However, no entry in the `Aces` list in the SharpHound output is created for the principal and therefore no AddSelf edge (or any other edges). 
![image](https://github.com/BloodHoundAD/SharpHoundCommon/assets/12843299/d4b05290-49d5-4d80-b51b-588b5cc50074)

[...]

PR #71, opened 2023-09-13
Fix BUG #70

godylockz:
## Description
See BUG #70 

## Motivation and Context
Properly make the "AddSelf" edge

## How Has This Been Tested?
Tested on a OU that has the AddSelf edge is created if you grant "Add/remove self as member" only, but if you select "All validated writes", then the edge is not created.

## Types of changes
The self ACL does not require the ACL for WriteMember.

### DIFF
--- a/src/CommonLib/Processors/ACLProcessor.cs
+++ b/src/CommonLib/Processors/ACLProcessor.cs
@@ -243,8 +243,7 @@ namespace SharpHoundCommonLib.Processors
                 //Cool ACE courtesy of @rookuu. Allows a principal to add itself to a group and no one else
                 if (aceRights.HasFlag(ActiveDirectoryRights.Self) &&
                     !aceRights.HasFlag(ActiveDirectoryRights.WriteProperty) &&
-                    !aceRights.HasFlag(ActiveDirectoryRights.GenericWrite) && objectType == Label.Group &&
-                    aceType == ACEGuids.WriteMember)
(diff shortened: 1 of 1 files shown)


godylockz:

JonasBK:
Thanks for your contribution @godylockz! 

I think the fix could produce false positives if a principal has been granted one of the other validated writes such as [Validated-DNS-Host-Name](https://learn.microsoft.com/en-us/windows/win32/adschema/r-validated-dns-host-name).

Hence we should check if the principal has been granted [Self-Membership](https://learn.microsoft.com/en-us/windows/win32/adschema/r-self-membership) specifically or all validated writes.

But I think it would make sense to introduce a new edge for all validated writes instead. I will confirm that with the team and get back to you.

godylockz:
I will be awaiting your response, thank you!

In the meantime if anyone feels the need to use these new binaries, posting them here.
[SharpHound-SelfMembership.zip](https://github.com/BloodHoundAD/SharpHoundCommon/files/12602548/SharpHound-SelfMembership.zip)

JonasBK:
Hi @godylockz,

Sorry for having you waiting so long.

I have talked to the team and we have decided that it would make the most sense to implement a new edge type named AllValidatedWrites to cover the scenario where a principal has an ACE with `ActiveDirectoryRights : Self` and `ObjectType : 00000000-0000-0000-0000-000000000000` (All) on Groups, as that is a different from having the specific permission of AddSelf.

Would you be up for updating your PR to create that edge and create the necessary BloodHound change as well? I will be available to guide you :)

PR #83, opened 2023-11-16
chore: add CertServices to collection methods

urangel:
Adds the CertServices Collection method to the ResolvedCollectionMethod enum based on updates on how we are partitioning ADCS collection.

### DIFF
--- a/src/CommonLib/Enums/CollectionMethods.cs
+++ b/src/CommonLib/Enums/CollectionMethods.cs
@@ -27,8 +27,8 @@ namespace SharpHoundCommonLib.Enums
         CertServices = 1 << 18,
         LocalGroups = DCOM | RDP | LocalAdmin | PSRemote,
         ComputerOnly = LocalGroups | Session | UserRights | CARegistry | DCRegistry,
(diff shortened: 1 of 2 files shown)


src/CommonLib/Enums/CollectionMethods.cs
```diff
@@ -24,6 +24,7 @@ public enum ResolvedCollectionMethod
         UserRights = 1 << 15,
         CARegistry = 1 << 16,
         DCRegistry = 1 << 17,
+        CertServices = 1 << 18,
         LocalGroups = DCOM | RDP | LocalAdmin | PSRemote,
         ComputerOnly = LocalGroups | Session | UserRights | CARegistry | DCRegistry,
         DCOnly = ACL | Container | Group | ObjectProps | Trusts | GPOLocalGroup,
```
JonasBK: I think it should be added to DCOnly as it is only collecting from DC. Would add it to Default as well.

urangel: I have pushed up these changes 👍

JonasBK (approved):

JonasBK (approved):

rvazarkar (approved):

PR #84, opened 2023-11-27
fix: add antixss lib and escape cns for cert templates

rvazarkar:
Fixes CNs with special characters breaking stuff


JonasBK (approved):
Works like a charm!

PR #86, opened 2023-11-29
fix: use correct constants for Enterprise CA flags

JonasBK:
We used the wrong flags:
image

Now the correct ones:
image


rvazarkar (approved):
Assuming you validated this, it lgtm

PR #89, opened 2023-12-23
Fix authentication with DirectoryEntry

nurfed1:
Had a case where sharphound would return invalid username and password. 
To fix the issue I had to provide ldap credentials to DirectoryEntry similar as in the [GetDomain](https://github.com/BloodHoundAD/SharpHoundCommon/blob/v3/src/CommonLib/LDAPUtils.cs#L1140) function


nurfed1:

rvazarkar (approved):

rvazarkar:
Thanks for the MR, merging this in!

PR #90, opened 2023-12-23
Check if ResolveAccountName returns a value

nurfed1:
The GPOLocalGroup processor crashes when ResolveAccountName fails to resolve the member name and this stops processing for the GPO entirely.
One such case I've seen is when the group name has a format like PT0-%ComputerName%-LocalAdmin. 

Ideally it would be great if some process variables such as %ComputerName% and %DomainName% could be resolved by sharphound, but I don't see a clean way to implement this.
But at least we can prevent a crash and print a warning.


rvazarkar:
Thanks for the MR, I spent some time trying to track this down but couldn't repro it locally

rvazarkar (approved):

PR #97, opened 2024-01-17
Add dockerfile to facilitate build on any platform

definitelynotagoblin:
## Description

A Dockerfile will provide an image with the dependencies necessary to build SharpHoundCommon.

## Motivation and Context

Provides a consistent environment to build within regardless of platform and SDK versions installed.

## How Has This Been Tested?

Tested on two macbooks and a linux machine.

## Types of changes

## Checklist:

-   [x] Documentation updates are needed, and have been made accordingly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,8 @@ FROM mono:6.12.0
 # Install .NET SDK
 ENV DOTNET_VERSION=5.0
 
-RUN curl -sSL https://dot.net/v1/dotnet-install.sh | bash -s -- -Channel $DOTNET_VERSION -InstallDir /usr/share/dotnet \
+RUN curl -sSL https://dot.net/v1/dotnet-install.sh  \
+    | bash -s -- -Channel $DOTNET_VERSION -InstallDir /usr/share/dotnet \
     && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
     
 WORKDIR /build
@@ -14,4 +15,4 @@ CMD [ "dotnet", "build" ]
 # docker build -t shc-build . --no-cache
 
 ## Build solution (every time):
-# docker run --rm -v ".:/build" shc-build
+# docker run --rm -v .:/build shc-build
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin -Channel $DOTNET_VERSION -InstallDir /usr/share/dotnet \
+    && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+    
+WORKDIR /SharpHoundCommon
+
+CMD [ "dotnet", "build" ]
+
+## Build Docker image (one time):
+# docker build -t sharphoundcommon-build . --no-cache
```
computator, line 14: It looks like this is missing the `-v` volume mount

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+## Build solution (every time):
+# docker run --rm -v "$(pwd):/SharpHoundCommon" sharphoundcommon-build
```
computator: this works for the volume mount, but you should also be able to use `.` for the current directory, making the quotes unnecessary: `-v .:/SharpHoundCommon`

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM mono:6.12.0
+
+# Install .NET SDK
+ENV DOTNET_VERSION=5.0
+
+RUN curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin -Channel $DOTNET_VERSION -InstallDir /usr/share/dotnet \
```
computator: `sh` and `bash` have a `-s` option specifically for reading from stdin, so you could also do:
```sh
... | bash -s -- -Channel $DOTNET_VERSION ...
```

Dockerfile
computator: never mind, I misread this

computator (approved):

PR #99, opened 2024-02-01
Filter empty hosts in host resolution

rvazarkar:
Prevents resolving empty hosts as localhost, creating false positives


superlinkx (approved):
Looks good

PR #100, opened 2024-02-02
Fix LDAP backoff retry logic for serverdown to actually use new connections

rvazarkar:
The current implementation was re-using the cached, invalid connection repeatedly and not actually creating a new one. The new implementation takes multi-consumer into account, properly creates new connections, and cleans up some bad logic around retry implementation.

### DIFF
--- a/src/CommonLib/LDAPUtils.cs
+++ b/src/CommonLib/LDAPUtils.cs
@@ -64,7 +64,7 @@ namespace SharpHoundCommonLib
         private readonly ConcurrentDictionary<string, string> _netbiosCache = new();
         private readonly PortScanner _portScanner;
         private LDAPConfig _ldapConfig = new();
-        private readonly ManualResetEvent _manualResetEvent = new(false);
+        private readonly ManualResetEvent _connectionResetEvent = new(false);
         
 
         /// <summary>
@@ -886,7 +886,7 @@ namespace SharpHoundCommonLib
                     retryCount++;
 
                     //If we are the holders of the reset event, we need to do logic to reset the connection
-                    if (_manualResetEvent.Reset())
+                    if (_connectionResetEvent.Reset())
                     {
                         try
                         {
@@ -906,13 +906,13 @@ namespace SharpHoundCommonLib
                         }
                         finally
                         {
-                            _manualResetEvent.Set();
+                            _connectionResetEvent.Set();
                         }
                     }
                     else
                     {
                         //If someone else is holding the reset event, we want to just wait and then pull the newly created connection out of the cache
-                        _manualResetEvent.WaitOne();
+                        _connectionResetEvent.WaitOne();
                         conn = CreateNewConnection(domainName, globalCatalog);
                     }
                     
--- a/src/CommonLib/LDAPUtils.cs
+++ b/src/CommonLib/LDAPUtils.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.DirectoryServices;
 using System.DirectoryServices.ActiveDirectory;
 using System.DirectoryServices.Protocols;
+using System.IO.Compression;
 using System.Linq;
 using System.Net;
 using System.Net.Sockets;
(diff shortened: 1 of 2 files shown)


src/CommonLib/LDAPUtils.cs
```diff
@@ -979,99 +1018,15 @@ private LdapConnection CreateNewConnection(string domainName = null, bool global
...
+            return QueryLDAP(ldapFilter, scope, props, new CancellationToken(), domainName, includeAcl, showDeleted,
+                adsPath, globalCatalog, skipCache, throwException);
+        }
 
-                pageControl.Cookie = pageResponse.Cookie;
-            }
+        private static TimeSpan GetNextBackoff(int retryCount)
+        {
+            return TimeSpan.FromSeconds(Math.Min(
+                BackoffDelayMultiplier * (retryCount + 1) * retryCount,
```
ddlees: This formula is fine but usually you'll see backoff calculation as something like this:

``` c#
TimeSpan.FromSeconds(Math.Min(
  MinBackoffDelay.TotalSeconds * Math.Pow(BackoffDelayMultiplier, retryCount),
  MaxBackoffDelay.TotalSeconds));
```

I believe they call this a truncated exponential backoff. You might consider adopting it and gate retrying the connection exclusively on the maximum number of retry attempts. That way you can simplify gating your retry attempts on just one variable and conform to a standard way of doing things.

ddlees (approved):
The semaphore makes the connection retry logic a little hard to reason about but the comments really help and it sounds like y'all are planning on refactoring this to be more async.

I left one nit but you can take it or leave it.

src/CommonLib/LDAPUtils.cs
```diff
@@ -66,6 +65,7 @@ public class LDAPUtils : ILDAPUtils
         private readonly ConcurrentDictionary<string, string> _netbiosCache = new();
         private readonly PortScanner _portScanner;
         private LDAPConfig _ldapConfig = new();
+        private readonly SemaphoreSlim _semaphoreSlim = new(1, 1);
```
computator: I think this can be specified as just `new(1)` since we are not specifying any pre-locked items.

src/CommonLib/LDAPUtils.cs
```diff
@@ -874,38 +873,79 @@ public IEnumerable<ISearchResultEntry> QueryLDAP(LDAPQueryOptions options)
...
+                    /*A ServerDown exception indicates that our connection is no longer valid for one of many reasons.
+                    However, this function is generally called by multiple threads, so we need to be careful in recreating
+                    the connection. Using a semaphore, we can ensure that only one thread is actually recreating the connection
+                    while the other threads that hit the ServerDown exception simply wait. The initial caller will hold the semaphore
+                    and do a backoff delay before trying to make a new connection which will replace the existing connection in the 
+                    _ldapConnections cache. Other threads will retrieve the new connection from the cache instead of making a new one
+                    This minimizes overhead of new connections while still fixing our core problem.
+                    
+                    A CurrentCount of 0 indicates the semaphore is already held because c# semaphores are weird and backwards*/
+                    var isSemaphoreHeld = _semaphoreSlim.CurrentCount == 0;
```
computator: This state is checked before the Semaphore is acquired, so it would be subject to a race condition where another thread could acquire the semaphore first. If that happens, then we would sleep and recreate a new connection despite just having done so. Or for the other way around it would try to use the bad cached connection again.

It would be better to track who is in charge of recreating the connection inside the semaphore some way, or  maybe use a different primitive. I couldn't figure out which C# primitive could be used in the way I was looking for, but I am thinking of some construct something more like this pseudocode:
```js
// try to acquire sync primitive (atomic op). This is nonblocking
held = primitive.TryAcquire()
if (held) {
  // we are first
  conn = delayandreconnect()
  // let everyone else thru
  primitive.Release()
else {
  // someone else already has the primitive. wait for the other thread to complete
  primitive.Wait()
  // the above Wait normally blocks until released, but will also return
  // immediately if Release has been called by another thread in between
  // here and the above TryAcquire

  // at this point the other thread has done the reconnection
  conn = getexistingconnection()

  // release acquired primitive (may be unnecessary depending on type)
  primitive.Release()
}
```

computator: Oh, I meant to say that perhaps a `Mutex` or a `ManualResetEvent|ManualResetEventSlim` might be more like the above construct? It's hard to tell from the docs though and someone more familiar with C# might know better.

src/CommonLib/LDAPUtils.cs
rvazarkar: You have to explicitly specify MaxCount, otherwise there is none defined: https://learn.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim.-ctor?view=net-8.0#system-threading-semaphoreslim-ctor(system-int32)

src/CommonLib/LDAPUtils.cs
rvazarkar: I replaced the semaphore with a ManualResetEvent, I think you're right it suits this use case better

superlinkx (approved):

definitelynotagoblin (approved):

computator (approved):

PR #101, opened 2024-02-03
Add NTLMv1 flag on GPO

Hackndo:
Similar to https://github.com/BloodHoundAD/SharpHound3/pull/47

If GPO object forces LmCompatibilityLevel to be less than 3, then the computers it will be applied on will use NTLMv1 when authenticating.

This information seems very useful from an attacking perspective as authentication can be coerced and NTLMv1 hash cracked or relayed without MIC

(Also https://github.com/BloodHoundAD/SharpHound/pull/87 on SharpHound)

image

### DIFF
--- a/src/CommonLib/OutputTypes/GPO.cs
+++ b/src/CommonLib/OutputTypes/GPO.cs
@@ -2,5 +2,6 @@
 {
     public class GPO : OutputBase
     {
+        public bool NTLMv1Enabled { get; set; }
     }
 }
--- a/src/CommonLib/Processors/GPOLmCompatibilityLevelProcessor.cs
+++ b/src/CommonLib/Processors/GPOLmCompatibilityLevelProcessor.cs
@@ -1,16 +1,11 @@
 ﻿using System;
-using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.DirectoryServices.Protocols;
 using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
-using System.Xml.XPath;
 using Microsoft.Extensions.Logging;
-using SharpHoundCommonLib.Enums;
 using SharpHoundCommonLib.LDAPQueries;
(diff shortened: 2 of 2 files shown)


Hackndo:

rvazarkar:
The implementation seems fine, but I'm sort of debating if it makes more sense to just merge this in with the existing gpo object processor stuff. @definitelynotagoblin @ddlees do either of you have any feelings on this? It is doing a bit more than our normal object processor does, but generally our processors are loosely linked to a particular collection method, so I think this would probably fit in ObjectProps instead

JonasBK:
Hi @Hackndo,

Thank you for the two PRs - awesome work! 🙌

We have discussed internally and reached the conclusion that we would like to have two properties for this setting you collect:

- LM Compatibility Level Raw (lmcompatibilitylevelraw)
- LM Compatibility Level (lmcompatibilitylevel)

The raw property should hold the int value of the registry setting. The other one should hold a string with corresponding setting value i.e. one of these:
- Send LM & NTLM responses
- Send LM & NTLM - use NTLMv2 session security if negotiated
- Send NTLM responses only
- Send NTLMv2 responses only
- Send NTLMv2 responses only. Refuse LM
- Send NTLMv2 responses only. Refuse LM & NTLM

Are you interested in updating your PR to create those properties? If not, then we will merge your PRs in and we will make the changes on top such that your commits still end up in the git history.

Let me know what you think and thanks again for contributing!

Hackndo:
Hello,
I'll update my PR in a few days.
Thank you for your feedback
**Edit (2nd of August)**: There were major changes in LDAPUtils in v4. I'll need some time to wrap my head around this. :)

JonasBK:
Hi @Hackndo,

Just checking if you are still interested in making the update to the PR :)

If not, we can still will merge your PRs in and make the changes on top such that your commits still end up in the git history.

Hackndo:
Hey there. I still have this in mind but yes, maybe it will be better if you merge and add the necessary changes on top.
Thanks 👌

PR #103, opened 2024-02-06
feat: GenericWrite on ADCS nodes

JonasBK:
Added GenericWrite for ADCS nodes.

We want to collect that for ESC4 and ESC5.

Ticket: BED-4084


rvazarkar (approved):
Approved as we have no outbound edges from certtemplates that could cause false positives in pathfinding

PR #108, opened 2024-03-12
Split ResolveHostToSid

definitelynotagoblin:
## Description

ResolveHostToSid resolves to a SID when a SID can be resolved, and falls back to a hostname from DNS when a SID cannot be resolved.  This produces unwanted behavior when a SID is expected but a hostname is returned, so we split the method in two - one that returns null where no SID can be resolved and another that falls back onto the hostname.

## Motivation and Context

https://specterops.atlassian.net/browse/BP-508

Bug in SHS when an AD object is deleted but remains on DNS.  Such objects currently are collected and recorded into our BloodHound graphs as empty domain CAs.

## Types of changes

## Checklist:

### DIFF
--- a/src/CommonLib/Helpers.cs
+++ b/src/CommonLib/Helpers.cs
@@ -327,6 +327,11 @@ namespace SharpHoundCommonLib
             CommonOids.SmartcardLogon,
             CommonOids.AnyPurpose
         };
+
+        public static bool IsSid(string sidContender)
+        {
+            return sidContender != null && sidContender.StartsWith("S-1-");
+        }
     }
 
     public class ParsedGPLink
--- a/src/CommonLib/LDAPUtils.cs
+++ b/src/CommonLib/LDAPUtils.cs
@@ -573,6 +573,38 @@ namespace SharpHoundCommonLib
         }
 
         /// <summary>
+        /// Takes a host in most applicable forms from AD and attempts to resolve it into a SID, falling back to hostname if SID cannot be resolved.
+        /// </summary>
+        /// <param name="hostname"></param>
+        /// <param name="domain"></param>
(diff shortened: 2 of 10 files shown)


definitelynotagoblin:
This will need to be merged with appropriate changes to SharpHound, SHS, and any other project that uses ResolveHostToSid from this library.

src/CommonLib/LDAPUtils.cs
```diff
@@ -572,6 +572,38 @@ public IEnumerable<string> DoRangedRetrieval(string distinguishedName, string at
             }
         }
 
+        /// <summary>
+        /// Takes a host in most applicable forms from AD and attempts to resolve it into a SID, falling back to hostname if SID cannot be resolved.
+        /// </summary>
+        /// <param name="hostname"></param>
+        /// <param name="domain"></param>
+        /// <returns></returns>
+        public async Task<string> ResolveHostToSidWithHostnameFallback(string hostname, string domain)
```
definitelynotagoblin, line 581: Here's the meat and potatoes.  Split ResolveHostToSid into two functions, one with and one without hostname resolution fallback.

src/CommonLib/Helpers.cs
```diff
@@ -327,6 +327,11 @@ public static IRegistryKey OpenRemoteRegistry(string target)
             CommonOids.SmartcardLogon,
             CommonOids.AnyPurpose
         };
+
+        public static bool IsSid(string sidContender)
+        {
+            return sidContender != null && sidContender.StartsWith("S-1-");
```
definitelynotagoblin, line 333: This isn't really that useful anymore.  The logic really isn't that prolific, and the places it's currently used can be replaced with simple 'not null' checks.

PR #127, opened 2024-05-21
LDAP Connection Logic Rewrite

rvazarkar:
## Description
Completely alters the way LDAP connections are made to be more stable

## Motivation and Context
LDAP connections are currently brittle and prone to breaking. Additionally, the GetDomain call is very flaky and can cause failures where it shouldn't. With this change, we use the following workflow:

```
/// Step 1: If domain passed in is non-null, skip this step
        // - Call GetDomain with a null domain to get the user's current domain
        // Step 2: Take domain passed in to the function or resolved from step 1
        // - Try an ldap connection on SSL
        // - If ServerUnavailable - Try an ldap connection on non-SSL
        //     Step 3: Pass the domain to GetDomain to resolve to a better name (potentially)
        //     - If we get a better name, repeat step 2 with the new name
        //     Step 4:
        // - Use GetDomain to get a domain object along with a list of domain controllers
        // - Try the primary domain controller on both ssl/non-ssl
        // - Loop over domain controllers and try each on ssl/non-ssl
```

This allows several fallback steps when negotiation ldap connections, which should significantly improve the ability of SharpHound to successfully connect

## How Has This Been Tested?

Local tests pass + testing in local environment. Additional tests incoming

## Types of changes

## Checklist:


definitelynotagoblin (approved):

src/CommonLib/LDAPConfig.cs
```diff
@@ -8,14 +8,25 @@ public class LDAPConfig
         public string Password { get; set; } = null;
         public string Server { get; set; } = null;
         public int Port { get; set; } = 0;
-        public bool SSL { get; set; } = false;
+        public bool ForceSSL { get; set; } = false;
```
zinic, line 11: Will this change the name of the configuration field for this in the JSON?

src/CommonLib/LDAPUtils.cs
```diff
@@ -483,22 +507,24 @@ public string GetSidFromDomainName(string domainName)
...
-                conn = task.ConfigureAwait(false).GetAwaiter().GetResult();
+                connWrapper = task.ConfigureAwait(false).GetAwaiter().GetResult();
             }
             catch
             {
                 yield break;
             }
 
-            if (conn == null)
+            if (connWrapper == null)
```
zinic: Is this an error condition if the connection wrapper is `null`?

src/CommonLib/LDAPUtils.cs
```diff
@@ -1078,6 +1102,11 @@ public ActiveDirectorySecurityDescriptor MakeSecurityDescriptor()
 
         public string BuildLdapPath(string dnPath, string domainName)
         {
+            //Check our cached info for a fast check
+            if (CachedDomainInfo.TryGetValue(domainName, out var info))
+            {
+                return $"{dnPath},{info.DomainSearchBase}";
+            }
             var domain = GetDomain(domainName)?.Name;
             if (domain == null)
```
zinic, line 1115: This feels a little awkward since the return type of this function is `string`. Why return `null` here?

src/CommonLib/LDAPUtils.cs
```diff
@@ -1415,163 +1438,534 @@ private static bool RequestNETBIOSNameFromComputer(string server, string domain,
...
+                    CachedDomainInfo.TryGetValue(domain, out info);
+                }
+                return new LdapConnectionWrapper
+                {
+                    Connection = connection,
+                    DomainInfo = info
+                };
+            }
+            
+            CheckAndThrowException(connectionResult.Exception);
```
zinic, line 1782: Should this be moved up higher in the function? There's a lot of logic between acquiring `connectionResult` and checking the `Exception` field.

src/CommonLib/LDAPUtils.cs
```diff
@@ -1415,163 +1438,534 @@ private static bool RequestNETBIOSNameFromComputer(string server, string domain,
...
+                }, fullServerName);
+            }
+            catch (LdapException e)
             {
-                ldapConnection.Dispose();
-                return connection;
-            });
+                try
+                {
+                    connection.Dispose();
```
zinic, line 1839: No action needed here, just thinking about the logic and flow here. There are a lot of calls in this logic to `Dispose` - we may want to, in the future, consider inverting control using a closure to encapsulate the logic so that `Dispose` is called only once in the function that receives the closure.

```
public TestResult TestConnection(SomeFunc testFunc)  {
    open_connection = OpenConnection(...)

    try {
        testFunc(connection)
    } finally {
        open_connection.Dispose()
    }
}
```

src/CommonLib/LDAPUtils.cs
```diff
@@ -1415,163 +1438,534 @@ private static bool RequestNETBIOSNameFromComputer(string server, string domain,
...
 
-        private async Task<string> GetUsableDomainController(Domain domain, bool gc = false)
+        private string GetDomainNetbiosName(LdapConnection connection, DomainInfo info)
         {
-            if (!gc && _domainControllerCache.TryGetValue(domain.Name.ToUpper(), out var dc))
-                return dc;
+            try
+            {
+                var searchRequest = new SearchRequest($"CN=Partitions,{info.DomainConfigurationPath}",
+                    "(&(nETBIOSName=*)(dnsRoot=*))",
```
zinic, line 1870: Kind request for documentation of this LDAP filter as they tend to be terse.

src/CommonLib/LDAPUtils.cs
```diff
@@ -1415,163 +1438,534 @@ private static bool RequestNETBIOSNameFromComputer(string server, string domain,
...
 
-            //If the PDC isn't reachable loop through the rest
-            foreach (DomainController domainController in domain.DomainControllers)
+        private string GetDomainSid(LdapConnection connection, DomainInfo info)
+        {
+            try
+            {
+                //This ldap filter searches for domain controllers
+                var searchRequest = new SearchRequest(info.DomainSearchBase,
+                    "(userAccountControl:1.2.840.113556.1.4.803:=8192)",
```
zinic: Kind request for documentation of this LDAP filter as they tend to be terse.

src/CommonLib/LDAPUtils.cs
```diff
@@ -1415,163 +1438,534 @@ private static bool RequestNETBIOSNameFromComputer(string server, string domain,
...
-                var name = domainController.Name;
-                if (!await _portScanner.CheckPort(name, port)) continue;
-                _log.LogInformation("Found usable Domain Controller for {Domain} : {PDC}", domain.Name, name);
-                _domainControllerCache.TryAdd(domain.Name.ToUpper(), name);
-                return name;
+                _log.LogWarning("Failed grabbing domainsid from ldap for {domain}", info.DomainFQDN);
+                return "";
             }
+        }
+        //
```
zinic: Rip dead comments or add context.

zinic (changes requested):
Got a few things that may or may not be issues and some thoughts.

src/CommonLib/LDAPConfig.cs
rvazarkar: We don't save this configuration anywhere, so its not an issue. This is a purely runtime config

src/CommonLib/LDAPUtils.cs
rvazarkar: Good catch, this should actually be checking the connection inside the wrapper, as the wrapper can never be null

src/CommonLib/LDAPUtils.cs
rvazarkar: We return null out of functions in most places in this codebase, so this is just consistent. We can talk about replacing them all with default returns instead, but null is an easy sentinel for "we couldn't figure this out"

src/CommonLib/LDAPUtils.cs
rvazarkar: The in between logic will only apply if we have a successful connection, and if we do we return before hitting this logic. It doesn't make any functional difference to the logic of this function, since success conditions never generate exceptions

src/CommonLib/LDAPUtils.cs
rvazarkar: We're specifically disposing of the connection only in error cases here, so its not quite so straightforward as just wrapping it and disposing

src/CommonLib/LDAPUtils.cs
zinic: Understood. Resolving this.

src/CommonLib/LDAPUtils.cs
zinic: Fair point!

src/CommonLib/LDAPUtils.cs
zinic: :+1:  Resolving this.

zinic (approved):
Changes this round look good

juggernot325 (approved):
I have limited C# knowledge, but the LDAP connection and error handling logic looks sound.

definitelynotagoblin (approved):

PR #135, opened 2024-07-19
IDirectoryObject Refactor

rvazarkar:
## Description
Replaces old interfaces with new IDirectoryObject interface for testability and cleanliness. Adds a whole bunch of tests around core logic. Also standardizes casing for LDAP to Ldap

## How Has This Been Tested?
Tested locally w/ testing + Lab VM. Merging into staging, not into main

## Types of changes

## Checklist:


definitelynotagoblin (approved):

src/CommonLib/DirectoryObjects/DirectoryEntryWrapper.cs
```diff
@@ -0,0 +1,188 @@
...
+        if (!CheckCache(propertyName)) {
+            return false;
+        }
+
+        var dest = new List<string>();
+        foreach (var val in _entry.Properties[propertyName]) {
+            if (val is string s) {
+                dest.Add(s);
+            }
+        }
```
computator, line 72: The other methods here return false if they don't match the type (string, Byte array, etc). Should this also do some type checking?

Also, is it correct to just drop array entries that are not strings?

src/CommonLib/DirectoryObjects/DirectoryEntryWrapper.cs
```diff
@@ -0,0 +1,188 @@
...
+        if (raw is not byte[][] b) {
+            return false;
+        }
+        value = b;
+        return true;
+    }
+
+    public bool TryGetIntProperty(string propertyName, out int value) {
+        value = 0;
+        if (!CheckCache(propertyName)) return false;
```
computator, line 94: Is this call redundant since `TryGetProperty` will immediately do the same thing anyway?

src/CommonLib/DirectoryObjects/DirectoryEntryWrapper.cs
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    public bool TryGetProperty(string propertyName, out string value) {
+        value = string.Empty;
+        if (!CheckCache(propertyName)) {
+            return false;
+        }
+
+        var s = _entry.Properties[propertyName][0];
```
computator: I see the other methods use `.Value` instead of `[0]`. I just wanted to verify this is correct?

src/CommonLib/DirectoryObjects/DirectoryEntryWrapper.cs
rvazarkar: This is intentional, we don't deal with non-string array types anywhere currently.

src/CommonLib/DirectoryObjects/DirectoryEntryWrapper.cs
rvazarkar: One returns an int, one returns a string, so other than that, kind of. The interface demands an implementation for it though

src/CommonLib/DirectoryObjects/DirectoryEntryWrapper.cs
rvazarkar: Good catch

src/CommonLib/OutputTypes/Computer.cs
```diff
@@ -35,7 +35,7 @@ public class ComputerStatus
         public string Error { get; set; }
 
         public static string NonWindowsOS => "NonWindowsOS";
-        public static string OldPwd => "PwdLastSetOutOfRange";
```
zinic, line 38: This an expected deletion? Looks like a constant and wanted to make sure.

src/CommonLib/Processors/ComputerAvailability.cs
```diff
@@ -89,23 +91,24 @@ public Task<ComputerStatus> IsComputerAvailable(ResolvedSearchResult result, ISe
             if (!_skipPasswordCheck)
             {
                 var passwordLastSet = Helpers.ConvertLdapTimeToLong(pwdLastSet);
+                var lastLogonTimeStamp = Helpers.ConvertLdapTimeToLong(lastLogon);
                 var threshold = DateTime.Now.AddDays(_computerExpiryDays * -1).ToFileTimeUtc();
 
-                if (passwordLastSet < threshold)
+                if (passwordLastSet < threshold && lastLogonTimeStamp < threshold)
```
zinic: All of the `passwordLastSet` logic may be better served and documented by extracting it into a method.

src/CommonLib/LdapProducerQueryGenerator.cs
```diff
@@ -0,0 +1,151 @@
...
+
+public class LdapProducerQueryGenerator {
+    public static GeneratedLdapParameters GenerateDefaultPartitionParameters(CollectionMethod methods) {
+        var filter = new LdapFilter();
+        var properties = new List<string>();
+        
+        properties.AddRange(CommonProperties.BaseQueryProps);
+        properties.AddRange(CommonProperties.TypeResolutionProps);
+
+        if (methods.HasFlag(CollectionMethod.ObjectProps) || methods.HasFlag(CollectionMethod.ACL) || methods.HasFlag(CollectionMethod.Container)) {
```
zinic, line 16: Almost feels like there's a pattern here to exercise but every time I think about trying to simplify this code I come up with something more complex than the boilerplate here. No change, just pontificating.

src/CommonLib/Processors/DomainTrustProcessor.cs
```diff
@@ -58,33 +56,26 @@ public async IAsyncEnumerable<DomainTrust> EnumerateDomainTrusts(string domain)
...
-                }
-                else
-                {
+                if (!entry.TryGetIntProperty(LDAPProperties.TrustDirection, out var td)) {
                     _log.LogTrace("Failed to convert trustdirection for target: {Domain}", domain);
                     continue;
                 }
 
-
+                trust.TrustDirection = (TrustDirection) td;
```
zinic, line 64: Are these casts safe?

test/unit/CertAbuseProcessorTest.cs
```diff
@@ -1,31 +1,19 @@
...
-using SharpHoundCommonLib;
-using SharpHoundCommonLib.Enums;
-using SharpHoundCommonLib.Processors;
-using Xunit;
 using Xunit.Abstractions;
 
-namespace CommonLibTest
-{
+namespace CommonLibTest {
     //TODO: Make these tests work
```
zinic, line 5: Can we get rid of this TODO?

test/unit/GroupProcessorTest.cs
zinic, line 1: This file has a bunch of warnings and it was touched in this changeset. Any possibility of patching up these broken windows?

test/unit/Utils.cs
```diff
@@ -24,15 +24,15 @@ internal static string B64ToString(string base64)
...
-        internal static async Task<T[]> ToArrayAsync<T>(this IAsyncEnumerable<T> items,
-            CancellationToken cancellationToken = default)
-        {
-            var results = new List<T>();
-            await foreach (var item in items.WithCancellation(cancellationToken)
-                               .ConfigureAwait(false))
-                results.Add(item);
-            return results.ToArray();
-        }
+        // internal static async Task<T[]> ToArrayAsync<T>(this IAsyncEnumerable<T> items,
```
zinic: If this is dead, let's pull it. We can always reference prior commits.

test/unit/WellKnownPrincipalTest.cs
zinic, line 1: Some easy nits to clean up here regarding unused variables.

zinic (changes requested):
Honker but good. Lots of excellent refactors and testing. Few minor nits and thoughts. I'd request a review of comments to see if anything is easy to fix before giving this formal approval.

src/CommonLib/OutputTypes/Computer.cs
rvazarkar: It got renamed to NotActive, since we're not checking more than just PwdLastSet

src/CommonLib/Processors/DomainTrustProcessor.cs
rvazarkar: Yes, int enum casts are always safe in c#

test/unit/CertAbuseProcessorTest.cs
rvazarkar: We didn't have time to fix up these tests so I'd like to leave them as a reminder for post-merge to get around to these

zinic (approved):
Shipit

src/CommonLib/DirectoryObjects/DirectoryEntryWrapper.cs
computator: I guess GitHub doesn't really highlight the lines I was commenting on very well. I was referring to the call to `CheckCache` on :94 which is also called via `TryGetProperty` on :96

src/CommonLib/LdapProducerQueryGenerator.cs
computator: I was thinking the same thing. It turns out these are bitfields tho, so what about this?
```suggestion
        if (methods & (CollectionMethod.ObjectProps | CollectionMethod.ACL | CollectionMethod.Container) != 0) {
```

Edit: Ah, maybe you were referring to all the code below and not just the if statement? In which case, what about the C# equivalent of this pseudocode?
```
checks = Dictionary<CollectionMethod,CommonProperties>(
    CollectionMethod.Container => CommonProperties.ContainerProps,
    CollectionMethod.Group => CommonProperties.GroupResolutionProps,
    ...
);
foreach (checks as method, prop)
    if (methods.HasFlag(method))
        properties.AddRange(prop);
```

PR #141, opened 2024-07-25
Bed-4487: fix handling for unicode control characters

mistahj67:
## Description

Closes BED-4487

Add protection around control characters such as null unicode `\u0000`by base64 encoding any properties we encounter.

## Motivation and Context

Running the collection with `--collectallproperties` brings in properties with binary control characters that cause issues downstream during ingestion into db's.

## How Has This Been Tested?

Added unit tests
Tested on AD lab

## Types of changes

## Checklist:

### DIFF
--- a/src/CommonLib/Processors/LdapPropertyProcessor.cs
+++ b/src/CommonLib/Processors/LdapPropertyProcessor.cs
@@ -15,13 +15,22 @@ using SharpHoundCommonLib.OutputTypes;
 
 namespace SharpHoundCommonLib.Processors {
     public class LdapPropertyProcessor {
-        private static readonly HashSet<string> ReservedAttributes = new HashSet<string>(CommonProperties.TypeResolutionProps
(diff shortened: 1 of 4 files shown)


src/CommonLib/Processors/LdapPropertyProcessor.cs
```diff
@@ -520,8 +520,16 @@ public class LdapPropertyProcessor {
         /// <param name="entry"></param>
         public Dictionary<string, object> ParseAllProperties(IDirectoryObject entry) {
             var props = new Dictionary<string, object>();
+
+            var type = typeof(LDAPProperties);
```
rvazarkar: This is going to revert a change we made in the newest commits. Instead, you'll need to add dsasignature to the ReservedAttributes list above and then use that.

rvazarkar (approved):

juggernot325 (approved):

PR #147, opened 2024-07-30
More fixes

rvazarkar:
## Description
A check was using > 0 instead of >= 0 which prevented the NTAuthCertificates node from being collected properly. Also adds some serialization tags to a directory object class for further testing

Preserves non-resolved ACEs in enterprise CA properties

Fixes a missing attribute store in generation

Uses `long` instead of `int` to allow fetching properties like LAPS expiration properly

Fixes the ldap filter for containers to exclude GPOs since they have the same objectclass

## Motivation and Context
Data inconsistency

## How Has This Been Tested?
Tests written

## Types of changes

## Checklist:

### DIFF
--- a/src/CommonLib/DirectoryObjects/DirectoryObjectExtensions.cs
+++ b/src/CommonLib/DirectoryObjects/DirectoryObjectExtensions.cs
@@ -35,7 +35,7 @@ public static class DirectoryObjectExtensions {
             return false;
         }
         
-        if (!directoryObject.TryGetIntProperty(LDAPProperties.Flags, out var flags)) {
+        if (!directoryObject.TryGetLongProperty(LDAPProperties.Flags, out var flags)) {
             flags = 0;
         }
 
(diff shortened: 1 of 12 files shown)


zinic (approved):
I have some nits but only related to unnamed string literals like the LDAP filter. Not worth visiting in this changeset but food for thought for the future.

Shipit

src/CommonLib/DirectoryObjects/DirectoryObjectExtensions.cs
```diff
@@ -35,15 +35,15 @@ public static class DirectoryObjectExtensions {
...
+        if (!directoryObject.TryGetLongProperty(LDAPProperties.Flags, out var flags)) {
             flags = 0;
         }
 
         directoryObject.TryGetDistinguishedName(out var distinguishedName);
         directoryObject.TryGetProperty(LDAPProperties.SAMAccountType, out var samAccountType);
         directoryObject.TryGetArrayProperty(LDAPProperties.ObjectClass, out var objectClasses);
 
-        return LdapUtils.ResolveLabel(objectIdentifier, distinguishedName, samAccountType, objectClasses, flags,
+        return LdapUtils.ResolveLabel(objectIdentifier, distinguishedName, samAccountType, objectClasses, (int)flags,
```
definitelynotagoblin, line 46: Do we want to keep this as an int?

definitelynotagoblin (approved):

src/CommonLib/DirectoryObjects/DirectoryObjectExtensions.cs
rvazarkar: Flags is always 0,1,or 2, so should be safe

PR #151, opened 2024-08-07
Semaphore fix

rvazarkar:
## Description
Our locking mechanism for ldap queries was being used incorrectly, leading to potential deadlocks and most likely general slowdowns. Move our query logic down into the connection pools and only lock the semaphore when a query is actively being performed

We also wrap all of our computer API calls in a timeout to ensure that queries that go out to lunch dont infinitely hang with a very generous 2 minute timeout by default

Additionally, up our default port scan timeout to 10 seconds. As the port scanner is checking asynchronously, this should not result in a significant performance loss as other coroutines will continue work. Testing discovered that the coroutine scheduler is not "fair" and that port scanner results varied significantly between `-c All` and `-c LoggedOn` for example.

## Types of changes

## Checklist:


zinic (approved):
This looks tight. Good find on the connection logic.

definitelynotagoblin (approved):

PR #161, opened 2024-09-13
[BED-4820] Fix: Min-valued nano durations would throw Overflow exceptions

definitelynotagoblin:
## Description

Min-value data for new duration properties throw OverflowException when converted to positive values via Math.Abs
While negative, the values collected actually represent positive numbers ...
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/e270cd0a-5295-41be-9e89-2c3dc3a39536?redirectedfrom=MSDN

It appears that min-value represents "N/A" data, or "Forever", so we'll record this value when resolving the property for this circumstance.

## Types of changes

## Checklist:

### DIFF
--- a/src/CommonLib/Processors/LdapPropertyProcessor.cs
+++ b/src/CommonLib/Processors/LdapPropertyProcessor.cs
@@ -724,6 +724,11 @@ namespace SharpHoundCommonLib.Processors {
 
         private static string ConvertNanoDuration(long duration)
         {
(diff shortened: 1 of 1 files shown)


JonasBK (approved):

rvazarkar (approved):

definitelynotagoblin:
Partial of https://github.com/BloodHoundAD/SharpHoundCommon/pull/162, merging the latter.

PR #176, opened 2024-12-02
BED-4965 - Owns/Owner Rework

Mayyhem:
## Description

Updated SharpHound to collect two new properties for every domain object:

- DoesAnyAceGrantOwnerRights
- DoesAnyInheritedAceGrantOwnerRights

## Motivation and Context

https://specterops.atlassian.net/browse/BED-4965

During ingest, BloodHound will use this information to determine whether any benign, non-abusable ACEs granted permissions to the OWNER RIGHTS SID (S-1-3-4), and whether or not any such ACEs were inherited. These properties allow ingest to determine whether any Owns and WriteOwner permissions granted are abusable or not.

## How Has This Been Tested?

Please refer to https://specterops.atlassian.net/wiki/spaces/BE/pages/750157858/Owns+WriteOwner for detailed testing information.

## Types of changes

## Checklist:

### DIFF
--- a/src/CommonLib/Processors/ACLProcessor.cs
+++ b/src/CommonLib/Processors/ACLProcessor.cs
@@ -2,10 +2,13 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.DirectoryServices;
+using System.Linq;
+using System.Net.Configuration;
 using System.Security.AccessControl;
 using System.Security.Cryptography;
 using System.Security.Principal;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using SharpHoundCommonLib.DirectoryObjects;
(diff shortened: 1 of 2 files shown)


src/CommonLib/Processors/ACLProcessor.cs
```diff
@@ -603,6 +612,435 @@ or Label.NTAuthStore
             }
         }
 
+        public async Task<(ACE[], bool, bool)> ProcessACL(byte[] ntSecurityDescriptor, string objectDomain,
```
definitelynotagoblin: Holy moly that's a function, but I don't see anything wrong here.

definitelynotagoblin (approved):

src/CommonLib/Processors/ACLProcessor.cs
```diff
@@ -244,365 +249,444 @@ public IEnumerable<string> GetInheritedAceHashes(byte[] ntSecurityDescriptor, st
...
+           Label objectType, bool hasLaps, string objectName = "")
+        {
+            var (aces, _, _) = await ProcessACL(ntSecurityDescriptor, objectDomain, objectType, hasLaps, objectName, true);
+            foreach (var ace in aces)
+            {
+                yield return ace;
+            }
+        }
+
+        public async Task<(ACE[], bool, bool)> ProcessACL(byte[] ntSecurityDescriptor, string objectDomain,
```
rvazarkar, line 261: Can we move the optional parameter to a non-optional one? Currently, the overload is actually only used if you manually specify the bool, since the other ProcessACL function has the same number of arguments. Let's make it explicit what we're asking for

rvazarkar: Its actually not going to build on CI either

Mayyhem: Thanks for the review, Rohan!

> Can we move the optional parameter to a non-optional one? Currently, the overload is actually only used if you manually specify the bool, since the other ProcessACL function has the same number of arguments. Let's make it explicit what we're asking for

I made `checkForOwnerRights` optional because `objectName` was already optional in the latest v4, and I couldn't add a required parameter after an optional one. Would you like me to make `objectName` required as well or split it into two methods, one for each signature?

> Its actually not going to build on CI either

What do I need to do to address this?

rvazarkar: No, just move `checkForOwnerRights` before `objectName`, since this is a new function it shouldn't matter

PR #177, opened 2024-12-09
Initial NTLM Implementation

rvazarkar:
## Description
This is the initial implementation of the new NTLM modelling, courtesy of @leechristensen . 

## Motivation and Context

https://specterops.atlassian.net/browse/BED-5113

## Types of changes

## Checklist:


src/CommonLib/Ntlm/HttpNtlmAuthenticationService.cs
```diff
@@ -0,0 +1,205 @@
...
+
+    public HttpNtlmAuthenticationService(ILogger logger, IHttpClientFactory httpClientFactory)
+    {
+        _logger = logger;
+        _httpClientFactory = httpClientFactory;
+    }
+
+    public async Task EnsureRequiresAuth(bool enableSigning, bool? useBadChannelBindings)
+    {
+        if (Url == null)
```
mvlipka: It looks like the `Url` property is only used on this method while being passed in to our other methods, might be worth moving this to a required argument rather than a property

src/CommonLib/Ntlm/HttpTransport.cs
```diff
@@ -0,0 +1,69 @@
...
+using System;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace SharpHoundCommonLib.Ntlm;
+
+public class HttpTransport : INtlmTransport
+{
+    private readonly ILogger _logger;
```
mvlipka, line 14: `_logger` doesn't seem to be used, might not be needed, or potentially was forgot to be used

src/CommonLib/Ntlm/HttpTransport.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        var response = await _httpClient.SendAsync(requestMessage);
+
+        if (!response.Headers.Contains("WWW-Authenticate"))
+        {
+            throw new InvalidOperationException("No WWW-Authenticate header found in response");
+        }
+
+        var authHeaders = response.Headers.WwwAuthenticate.Where(a => a.Scheme == _authScheme);
+        if (!authHeaders.Any())
```
mvlipka, line 38: Slight performance nit: I think we can use `authHeaders.Count == 0` here

src/CommonLib/Ntlm/HttpTransport.cs
```diff
@@ -0,0 +1,69 @@
...
+        requestMessage.Headers.Add("Authorization", $"{_authScheme} {messageBase64}");
+
+        var response = await _httpClient.SendAsync(requestMessage);
+
+        if (!response.Headers.Contains("WWW-Authenticate"))
+        {
+            throw new InvalidOperationException("No WWW-Authenticate header found in response");
+        }
+
+        var authHeaders = response.Headers.WwwAuthenticate.Where(a => a.Scheme == _authScheme);
```
mvlipka: Slight performance nit: we can convert this to a list to avoid enumerating the elements again, probably not a large enough collection to make a large difference

src/CommonLib/Ntlm/LdapNative.cs
```diff
@@ -0,0 +1,346 @@
...
+
+            // Freeing the memory always results in crashes. Is it managed by something else?
+            //if (response != IntPtr.Zero)
+            //{
+            //    var responseBerval = Marshal.PtrToStructure<NativeMethods.berval>(response);
+            //    if (responseBerval.bv_val != IntPtr.Zero)
+            //    {
+            //        Marshal.FreeHGlobal(responseBerval.bv_val);
+            //    }
+            //    Marshal.FreeHGlobal(response);
```
mvlipka, line 241: I _think_ that the memory is cleaned up at the end of the PInvoke call since, which could potentially be causing the crash

src/CommonLib/Ntlm/NtlmAuthenticationHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+                AuthenticationMethod.NTLM,
+                _targetService,
+                Options.Bindings,
+                Options.Signing,
+                Options.Signing
+            );
+
+        // NEGOTIATE
+        var negotiateMsgBytes = context.Step();
+        //_logger.LogDebug($"NTLM-NEGOTIATE message: {Convert.ToBase64String(negotiateMsgBytes)}");
```
mvlipka: Not sure if intentionally left out, but _logger is no longer used if intentional

We also don't use `_host` or `NtlmAuthPackageName`

src/CommonLib/OutputTypes/APIResult.cs
```diff
@@ -1,8 +1,52 @@
 ﻿namespace SharpHoundCommonLib.OutputTypes
 {
-    public class APIResult
+    public class BaseAPIResult<T>
```
mvlipka: Unused, can probably remove

src/CommonLib/OutputTypes/NtlmSession.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            return ($"  {TimeCreatedUtc?.ToLocalTime()},{Id},{targetUser},{AccountSid},{SourceHost},,{PackageName}");
+        }
+
+        public static NtlmSession FromLogonEvent(string host, EventRecord evnt)
+        {
+            if (evnt.Id != 4624)
+                throw new ArgumentException("Not a logon event");
+
+            //var subjectUserSid = eventDetail.Properties[0].Value.ToString();
```
mvlipka: Thinking we can remove these

src/CommonLib/OutputTypes/NtlmSession.cs
```diff
@@ -0,0 +1,113 @@
...
+        public string? AccountDomain { get; set; } = accountDomain;
+        public string SourceHost { get; set; } = sourceHost;   // The host the auth originated from
+        public string? SourceIp { get; set; } = sourceIp;
+        public string? SourcePort { get; set; } = sourcePort;
+        public string? PackageName { get; set; } = packageName;
+
+        public override string ToString()
+        {
+            var targetUser = AccountDomain + "\\" + AccountName;
+            var source = (SourceIp != null || SourcePort != null) ? $"{SourceIp}:{SourcePort}" : "";
```
mvlipka, line 36: Unused, maybe was meant for logging?

src/CommonLib/OutputTypes/NtlmSession.cs
```diff
@@ -0,0 +1,113 @@
...
+        string? accountSid,
+        string accountName,
+        string? accountDomain,
+        string sourceHost,
+        string? sourceIp,
+        string? sourcePort,
+        string packageName
+        )
+    {
+        public string Host { get; set; } = host;   // Hostname the sessions occurred on
```
mvlipka: Unused

src/CommonLib/OutputTypes/NtlmSession.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            var targetUser = AccountDomain + "\\" + AccountName;
+            var source = (SourceIp != null || SourcePort != null) ? $"{SourceIp}:{SourcePort}" : "";
+
+            return ($"  {TimeCreatedUtc?.ToLocalTime()},{Id},{targetUser},{AccountSid},{SourceHost},,{PackageName}");
+        }
+
+        public static NtlmSession FromLogonEvent(string host, EventRecord evnt)
+        {
+            if (evnt.Id != 4624)
```
mvlipka: Should these magic numbers be moved to an enum? Maybe a constant for this

src/CommonLib/Processors/CAEnrollmentProcessor.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        public async Task<IEnumerable<ApiResult<CAEnrollmentEndpoint>>> ScanAsync()
+        {
+            var endpoints = new List<ApiResult<CAEnrollmentEndpoint>>();
+
+            try
+            {
+                var webEnrollmentTask = ScanHttpEndpoint(CAEnrollmentEndpointType.WebEnrollmentApplication);
+                var webServiceTask = ScanHttpEndpoint(CAEnrollmentEndpointType.EnrollmentWebService);
+                // var rpcServiceTask = CheckEnrollmentRpcAsync();
```
mvlipka: Can remove the comments

src/CommonLib/Processors/DCLdapProcessor.cs
```diff
@@ -0,0 +1,291 @@
...
+
+            return ApiResult<bool>.CreateSuccess(accessibleWithoutSigning == false);
+        }
+        catch (Exception ex)
+        {
+            return ApiResult<bool>.CreateError($"CheckIsNtlmSigningRequired failed: {ex}");
+        }
+    }
+
+    // Checks if EPA is enabled. Does so by trying to auth with wrong channel bingings.
```
mvlipka: bingings->bindings

src/CommonLib/Processors/DCLdapProcessor.cs
```diff
@@ -0,0 +1,291 @@
...
+            };
+            var accessibleWithNoBindings = await Authenticate(_ldapSslEndpoint, new LdapAuthOptions()
+            {
+                Signing = false,
+                Bindings = bindings
+            });
+
+            return ApiResult<bool>.CreateSuccess(accessibleWithNoBindings == true);
+
+            //// Second, check if we can connect, but with *INVALID* bindings
```
mvlipka: Looks like we can remove these

src/CommonLib/Processors/DCLdapProcessor.cs
```diff
@@ -0,0 +1,291 @@
...
+        var auth = new NtlmAuthenticationHandler(_log, host, $"LDAP/{host.ToUpper()}")
+        {
+            Options = options
+        };
+        var transport = new LdapTransport(_log, endpoint);
+
+        try
+        {
+            transport.InitializeConnectionAsync(_ldapTimeout);
+            var bytes = (byte[])await auth.PerformNtlmAuthenticationAsync(transport);
```
mvlipka: bytes is unused, we can avoid the extra memory initialization

src/CommonLib/Ntlm/LdapNative.cs
```diff
@@ -0,0 +1,346 @@
...
+            // If anything goes wrong getting the server error,
+            // return null rather than throw since we're already
+            // in error handling code
+            return null;
+        }
+    }
+
+    public void CheckError(int result, string operation)
+    {
+        if (result != 0)
```
mvlipka: I think we could replace `0` with `LdapErrorCodes.Success`, but not sure if the return of the `ldap_set_option` system call is the same as the out value

src/CommonLib/Processors/DCLdapProcessor.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+        catch (Exception ex)
+        {
+            _log.LogError("An unhandled error occurred during the LDAP test: {ex}", ex);
+        }
+
+        return false;
+    }
+
+    //private Task<bool> CanAccessWithChannelBindings()
```
mvlipka: Removable

src/CommonLib/Processors/EventLogProcessor.cs
```diff
@@ -0,0 +1,332 @@
...
+         string host,
+         string domain,
+         EventLogCollection collectionType,
+         int numDays = 7,
+         int readEventDelayMs = 5,
+         int readEventTimeoutMs = 10000,
+         NetworkCredential? localAdminCredential = null
+         )
+    {
+        private readonly ILdapUtils _utils = utils;
```
mvlipka: Looks like a majority of these private fields are unused

src/CommonLib/Processors/EventLogProcessor.cs
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        private List<NtlmSession> CollectEvents(string query)
+        {
+            var ntlmSessions = new List<NtlmSession>();
+            var dupCheck = new HashSet<string>();
+
+            using EventLogReader logReader = GetEventLogReader(_host, "Security", query);
+
+            int count = 0;
```
mvlipka: We don't seem to do anything with the `count` here

src/CommonLib/Processors/WebClientServiceProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Threading.Tasks;
+
+namespace SharpHoundCommonLib.Processors
+{
+    public class WebClientServiceProcessor(ILogger log = null)
+    {
+        private readonly ILogger _log = log ?? Logging.LogProvider.CreateLogger("WebClientServiceProcessor");
+
+        // Define constants
+        public const uint MAXIMUM_ALLOWED = 0x02000000;
```
mvlipka, line 21: I believe these match the naming style of the library calls, but not sure we want to copy that naming style into our managed code

src/CommonLib/Processors/WebClientServiceProcessor.cs
```diff
@@ -0,0 +1,99 @@
...
+            uint dwDesiredAccess,
+            uint dwShareMode,
+            IntPtr lpSecurityAttributes,
+            uint dwCreationDisposition,
+            uint dwFlagsAndAttributes,
+            IntPtr hTemplateFile);
+
+        public bool TestPathExists(string path)
+        {
+            using SafeFileHandle handle = CreateFile(
```
mvlipka: Is there a benefit of this over using `File.Create`?

src/CommonLib/ThirdParty/PSOpenAD/Authentication.cs
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    public override byte[] Wrap(ReadOnlySpan<byte> data, bool encrypt) => Array.Empty<byte>();
+    public override byte[] Unwrap(ReadOnlySpan<byte> data) => Array.Empty<byte>();
+    public override uint MaxWrapSize(uint outputSize, bool confReq) => 0;
+    public override void Dispose()
+    { }
+}
+
+//internal class GssapiContext : SecurityContext
```
mvlipka: Removable

src/CommonLib/ThirdParty/PSOpenAD/Authentication.cs
```diff
@@ -0,0 +1,515 @@
...
+    private UInt32 _trailerSize = 0;
+    private UInt32 _seqNo = 0;
+
+    public SspiContext(string? username, string? password, AuthenticationMethod method, string target,
+        ChannelBindings? channelBindings, bool integrity, bool confidentiality)
+    {
+        _bindingData = CreateChannelBindings(channelBindings);
+        _targetSpn = target;
+
+        string package = Enum.GetName(typeof(AuthenticationMethod), method);
```
mvlipka: I think `method.ToString()` might be more appropriate here

src/CommonLib/ThirdParty/PSOpenAD/Authentication.cs
```diff
@@ -0,0 +1,515 @@
...
+
+                if (_bindingData != null)
+                {
+                    inputBuffers[idx].cbBuffer = (UInt32)_bindingData.Length;
+                    inputBuffers[idx].BufferType = (UInt32)SecBufferType.SECBUFFER_CHANNEL_BINDINGS;
+                    inputBuffers[idx].pvBuffer = (IntPtr)cbBuffer;
+                }
+
+                SspiSecContext context = SSPI.InitializeSecurityContext(_credential, _context, _targetSpn, _flags,
+                    TargetDataRep.SECURITY_NATIVE_DREP, inputBuffers, new[] { SecBufferType.SECBUFFER_TOKEN, });
```
mvlipka, line 220: Nit: we can replace `new [] {}` with `[SecBufferType.SECBUFFER_TOKEN]`

src/CommonLib/ThirdParty/PSOpenAD/Authentication.cs
```diff
@@ -0,0 +1,515 @@
...
+                        offset += (int)buffers[0].cbBuffer;
+                    }
+
+                    Marshal.Copy((IntPtr)dataPtr, wrapped, offset, (int)buffers[1].cbBuffer);
+                    offset += (int)buffers[1].cbBuffer;
+
+                    if (buffers[2].cbBuffer > 0)
+                    {
+                        Buffer.BlockCopy(padding, 0, wrapped, offset, (int)buffers[2].cbBuffer);
+                        offset += (int)buffers[2].cbBuffer;
```
mvlipka, line 292: We can remove since we no longer use the `offset` after this if block, might have been kept for future changes, though?

src/CommonLib/ThirdParty/PSOpenAD/SSPI.cs
```diff
@@ -0,0 +1,676 @@
...
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+using System.Security.Authentication;
+
+namespace SharpHoundCommonLib.ThirdParty.PSOpenAD;
+
+internal static partial class Helpers
+{
+    [StructLayout(LayoutKind.Sequential)]
+    public struct SEC_CHANNEL_BINDINGS
```
mvlipka, line 15: Pretty sure these are the names from the shared library, but we should follow our managed code's styling

src/CommonLib/ThirdParty/PSOpenAD/SSPI.cs
```diff
@@ -0,0 +1,676 @@
...
+        OutputBuffers = outputBuffers;
+        Expiry = expiry;
+        Flags = flags;
+        MoreNeeded = moreNeeded;
+    }
+}
+
+internal static class SSPI
+{
+    private const Int32 SEC_I_CONTINUE_NEEDED = 0x00090312;
```
mvlipka, line 126: Same here with matching C#'s styles

src/CommonLib/Ntlm/LdapNative.cs
```diff
@@ -0,0 +1,346 @@
...
+
+    [DllImport("Wldap32.dll", EntryPoint = "ldap_sasl_bind_s", CallingConvention = CallingConvention.Cdecl)]
+    public static extern int ldap_sasl_bind_s(
+        SafeLdapHandle ld,
+        string dn,
+        string mechanism,
+        IntPtr cred,
+        IntPtr serverctrls,
+        IntPtr clientctrls,
+        out IntPtr msgidp);
```
mvlipka, line 297: Nit: this name doesn't match the output defined in the library
https://learn.microsoft.com/en-us/windows/win32/api/winldap/nf-winldap-ldap_sasl_bind_sa

src/CommonLib/ThirdParty/PSOpenAD/SSPI.cs
```diff
@@ -0,0 +1,676 @@
...
+    SECPKG_CRED_BOTH = 0x00000003,
+    SECPKG_CRED_DEFAULT = 0x00000004,
+    SECPKG_CRED_AUTOLOGON_RESTRICTED = 0x00000010,
+    SECPKG_CRED_PROCESS_POLICY_ONLY = 0x00000020,
+}
+
+[Flags]
+internal enum InitiatorContextRequestFlags : uint
+{
+    ISC_REQ_DELEGATE = 0x00000001,
```
mvlipka, line 481: Naming in these enums as well

mvlipka (changes requested):
Looking good, mostly style comments and naming conventions that I think we can improve on

src/CommonLib/LdapProducerQueryGenerator.cs
```diff
@@ -46,7 +46,7 @@ public static GeneratedLdapParameters GenerateDefaultPartitionParameters(Collect
             if (methods.HasFlag(CollectionMethod.SPNTargets))
                 properties.AddRange(CommonProperties.SPNTargetProps);
 
-            if (methods.HasFlag(CollectionMethod.DCRegistry))
+            if (methods.HasFlag(CollectionMethod.DCRegistry) || methods.HasFlag(CollectionMethod.LdapServices))
```
rvazarkar: Does this need WebClientService/SMBInfo/EventLogs added to it?

src/CommonLib/Ntlm/LdapNative.cs
rvazarkar: Unfortunately, pinvoke is weird, this is normal

src/CommonLib/Processors/WebClientServiceProcessor.cs
rvazarkar: Good question, @leechristensen can you clarify?

leechristensen: I believe it was because File.Create always throws an exception for the pipe whereas we can get a success (and more granular error checking) with lower-level win32 APIs. I believe the error handling seemed simpler for Win32 APIs vs dealing with the various exceptions that .NET's File class may throw. But not tied to it either way.

src/CommonLib/Ntlm/HttpTransport.cs
rvazarkar: I'm ok with leaving unused loggers for now, it lets us quickly add logging later

src/CommonLib/OutputTypes/NtlmSession.cs
rvazarkar: We can, but I think they serve as documentation for that structure thats coming back, which isn't bad to have in case we need the info later

src/CommonLib/Processors/WebClientServiceProcessor.cs
rvazarkar: I'm going to leave it as is to match the well defined consts

src/CommonLib/Processors/WebClientServiceProcessor.cs
rvazarkar: Adding a comment and resolving

src/CommonLib/ThirdParty/PSOpenAD/Authentication.cs
rvazarkar: I don't know the intention for that so I'm going to leave it as is for now, we can refactor later

rvazarkar:
recheck

src/CommonLib/OutputTypes/NtlmSession.cs
rvazarkar: Added into the log statement

mvlipka (approved):

src/CommonLib/LdapUtils.cs
```diff
@@ -1203,6 +1203,15 @@ await utils.GetDomainNameFromSid(objectIdentifier) is (true, var domainName)) {
                 return (true, res);
             }
 
+            res.ObjectType = await ComputeLabel(directoryObject, objectIdentifier, domain, utils);
+
+            directoryObject.TryGetProperty(LDAPProperties.SAMAccountName, out var samAccountName);
+            res.DisplayName = ComputeDisplayName(directoryObject, domain, res.ObjectType, samAccountName);
```
definitelynotagoblin, line 1210: I would be careful that this returns an expected DisplayName when samAccountName is `null`

definitelynotagoblin: If objecttype is `Base`, this may return something like "@[domain]"

rvazarkar: Added some fallbacks

src/CommonLib/Ntlm/HttpNtlmAuthenticationService.cs
```diff
@@ -0,0 +1,167 @@
...
+        _logger = logger ?? Logging.LogProvider.CreateLogger(nameof(HttpNtlmAuthenticationService));
+        _httpClientFactory = httpClientFactory;
+    }
+
+    public async Task EnsureRequiresAuth(Uri url, bool? useBadChannelBindings) {
+        if (url == null)
+            throw new ArgumentException("Url property is null");
+
+        if (useBadChannelBindings == null && url.Scheme == "https")
+            throw new ArgumentException("When using HTTPS, useBadChannelBindings must be set");
```
definitelynotagoblin, line 30: Nitpicking, but trinary args like this tend to create some funky logic flows that can be hard to follow if you're smooth-brained like I am.
Here I think we're taking on some complexity by managing arg validations that the callers could be doing instead.

I might split this into two functions for clarity: the public function that does everything up to this point, then a private `EnsureRequiresAuth(Uri url, bool useBadChannelBindings)` that this one calls, feeding the appropriate coerced `useBadChannelBindings`:

`EnsureRequiresAuth(url, useBadChannelBindings ?? true);`

src/CommonLib/Ntlm/HttpNtlmAuthenticationService.cs
```diff
@@ -0,0 +1,167 @@
...
+        var schemes = response.Headers.WwwAuthenticate
+            .Select(header => header.Scheme)
+            .Where(scheme => scheme == "NTLM" || scheme == "Negotiate")
+            .Distinct()
+            .ToArray();
+
+        return schemes;
+    }
+
+    private async Task AuthWithBadChannelBindings(Uri url, string authScheme) {
```
definitelynotagoblin: Could use "-Async" suffix

src/CommonLib/OutputTypes/APIResult/APIResult.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        public static APIResult<T> Failure(string failureReason) {
+            return new APIResult<T> {
+                Collected = false,
+                FailureReason = failureReason
+            };
+        }
+        
+        public static implicit operator APIResult<T>(T input)
```
definitelynotagoblin: Today I learned implicit operators

src/CommonLib/OutputTypes/APIResult/AceRegistryAPIResult.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace SharpHoundCommonLib.OutputTypes
 {
-    public class AceRegistryAPIResult : APIResult
+    public class AceRegistryAPIResult : APIResult.APIResult
```
definitelynotagoblin: Prefer the namespace and class had different names but not a blocker

src/CommonLib/OutputTypes/Computer.cs
```diff
@@ -1,37 +1,76 @@
...
     /// </summary>
-    public class Computer : OutputBase
-    {
+    public class Computer : OutputBase {
         public string PrimaryGroupSID { get; set; }
-        public TypedPrincipal[] AllowedToDelegate { get; set; } = Array.Empty<TypedPrincipal>();
-        public TypedPrincipal[] AllowedToAct { get; set; } = Array.Empty<TypedPrincipal>();
-        public TypedPrincipal[] HasSIDHistory { get; set; } = Array.Empty<TypedPrincipal>();
-        public TypedPrincipal[] DumpSMSAPassword { get; set; } = Array.Empty<TypedPrincipal>();
+        public TypedPrincipal[] AllowedToDelegate { get; set; } = [];
```
definitelynotagoblin, line 10: This syntax may not compile with older dotnet sdks

definitelynotagoblin: Looks like this syntax feature isn't available til dotnet 8
`Feature 'collection expressions' is not available in C# 11.0. Please use language version 12.0 or greater.`

definitelynotagoblin: Thanks dotnet fiddle

definitelynotagoblin (changes requested):
https://github.com/SpecterOps/SharpHoundCommon/pull/177/files#r1945552460

definitelynotagoblin (approved):

src/CommonLib/Enums/CollectionMethod.cs
```diff
@@ -25,10 +23,19 @@ public enum CollectionMethod
         CARegistry = 1 << 16,
         DCRegistry = 1 << 17,
         CertServices = 1 << 18,
+        LdapServices = 1 << 19,
+        WebClientService = 1 << 21,
+        SmbInfo = 1 << 22,
+        //TODO: Re-introduce this when we're ready for Event Log collection
+        //EventLogs = 1 << 23,
```
ktstrader, line 30: Looks like EventLogs should be uncommented because it is used in the ntlm updates for SharpHound [here](https://github.com/SpecterOps/SharpHound/pull/133/files#diff-81351f3d92bff48f24fe7b072073d3639d382eb32da99bfdbc200e6807fc2e99R210)

rvazarkar: SH is not doing event log collection yet, so remove the reference on that side as well

src/CommonLib/Processors/CertAbuseProcessor.cs
```diff
@@ -245,32 +237,26 @@ private RegistryResult GetEnrollmentAgentRights(string target, string caName)
...
             {
-                ret.FailureReason = data.FailureReason;
-                return ret;
+                return APIResult<bool>.Failure(data.FailureReason);
             }
 
             if (data.Value == null)
             {
-                return ret;
+                return APIResult<bool>.Success(false);
```
ktstrader, line 255: @leechristensen Changing type BoolRegistryAPIResult to APIResult<bool>, when data.Value is null I did have to set the return value to false. Is this an accurate value?

leechristensen: I believe that's correct, but I'll defer to @JonasBK since he's more familiar with this.

JonasBK: That is correct! 👍

PR #186, opened 2025-02-26
Add NTLM Registry Collection Processors

rvazarkar:
## Description
Adds the ability to collect specific registry keys from computers courtesy of @leechristensen

## Motivation and Context
Part of the NTLM effort

## Types of changes

## Checklist:

### DIFF
--- a/src/CommonLib/LdapUtils.cs
+++ b/src/CommonLib/LdapUtils.cs
@@ -19,6 +19,7 @@ using SharpHoundCommonLib.LDAPQueries;
 using SharpHoundCommonLib.OutputTypes;
 using SharpHoundCommonLib.Processors;
 using SharpHoundRPC.NetAPINative;
+using SharpHoundRPC.PortScanner;
 using Domain = System.DirectoryServices.ActiveDirectory.Domain;
 using Group = SharpHoundCommonLib.OutputTypes.Group;
 using SearchScope = System.DirectoryServices.Protocols.SearchScope;
@@ -43,7 +44,7 @@ namespace SharpHoundCommonLib {
             new(StringComparer.OrdinalIgnoreCase);
 
         private readonly ILogger _log;
-        private readonly PortScanner _portScanner;
+        private readonly IPortScanner _portScanner;
         private readonly NativeMethods _nativeMethods;
         private readonly string _nullCacheKey = Guid.NewGuid().ToString();
         private static readonly Regex SIDRegex = new(@"^(S-\d+-\d+-\d+-\d+-\d+-\d+)(-\d+)?$");
--- a/src/CommonLib/Processors/ComputerAvailability.cs
+++ b/src/CommonLib/Processors/ComputerAvailability.cs
@@ -25,7 +25,7 @@ namespace SharpHoundCommonLib.Processors {
             _skipPasswordCheck = skipPasswordCheck;
         }
 
-        public ComputerAvailability(PortScanner scanner, int timeout = 500, int computerExpiryDays = 60,
+        public ComputerAvailability(IPortScanner scanner, int timeout = 500, int computerExpiryDays = 60,
             bool skipPortScan = false, bool skipPasswordCheck = false,
             ILogger log = null) {
             _scanner = scanner ?? new PortScanner();
--- a/src/CommonLib/Processors/ComputerAvailability.cs
+++ b/src/CommonLib/Processors/ComputerAvailability.cs
@@ -2,23 +2,21 @@
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using SharpHoundCommonLib.OutputTypes;
+using SharpHoundRPC.PortScanner;
 
-namespace SharpHoundCommonLib.Processors
-{
-    public class ComputerAvailability
-    {
+namespace SharpHoundCommonLib.Processors {
+    public class ComputerAvailability {
         public delegate Task ComputerStatusDelegate(CSVComputerStatus status);
 
         private readonly int _computerExpiryDays;
         private readonly ILogger _log;
-        private readonly PortScanner _scanner;
+        private readonly IPortScanner _scanner;
         private readonly int _scanTimeout;
         private readonly bool _skipPasswordCheck;
         private readonly bool _skipPortScan;
 
         public ComputerAvailability(int timeout = 10000, int computerExpiryDays = 60, bool skipPortScan = false,
-            bool skipPasswordCheck = false, ILogger log = null)
-        {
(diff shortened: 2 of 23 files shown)


src/CommonLib/Processors/PortScanner.cs
```diff
@@ -28,75 +25,73 @@ public PortScanner(ILogger log = null)
...
+
                 PortScanCache.TryAdd(key, false);
                 return false;
-            }
-            catch (Exception e)
-            {
+            } catch (Exception e) {
                 _log.LogDebug(e, "Exception checking {Hostname}:{Port}", hostname, port);
+                if (throwError) {
+                    throw;
```
ktstrader, line 66: What is the point of a `throw` inside of a `catch` block?

src/CommonLib/Processors/ComputerAvailability.cs
```diff
@@ -29,9 +27,8 @@ public ComputerAvailability(int timeout = 10000, int computerExpiryDays = 60, bo
 
         public ComputerAvailability(PortScanner scanner, int timeout = 500, int computerExpiryDays = 60,
```
definitelynotagoblin: Should this take an `IPortScanner` param or do we need the concrete implementation for this one?

src/CommonLib/Processors/RegistryProcessor.cs
```diff
@@ -0,0 +1,95 @@
...
+                    output.RestrictSendingNtlmTraffic = Convert.ToUInt32(key.Value);
+
+                if (name == "RestrictReceivingNTLMTraffic")
+                    output.RestrictReceivingNtlmTraffic = Convert.ToUInt32(key.Value);
+
+                if (name == "LMCompatibilityLevel")
+                    output.LmCompatibilityLevel = Convert.ToUInt32(key.Value);
+
+                if (name == "UseMachineId")
+                    output.UseMachine = Convert.ToUInt32(key.Value);
```
ktstrader: Would it be cleaner if a switch block was used here instead?

src/CommonLib/Processors/PortScanner.cs
definitelynotagoblin: In this case we want the PortScanner to bubble exceptions up, but only if the `throwError` flag is set, otherwise log and suppress them

src/SharpHoundRPC/Registry/DotNetWmiRegistryStrategy.cs
```diff
@@ -0,0 +1,152 @@
...
+                    scope = new ManagementScope("root\\cimv2");
+                } else {
+                    var connectionOptions = new ConnectionOptions {
+                        Authority = UseKerberos
+                            ? @$"kerberos:{Domain}\{targetMachine}"
+                            : @$"ntlmdomain:{Domain}"
+                    };
+
+                    scope = new ManagementScope(
+                        $"\\\\{targetMachine}\\root\\cimv2",
```
definitelynotagoblin: Why not use the `@` string syntax as above to auto-escape the control chars?

src/SharpHoundRPC/Registry/DotNetWmiRegistryStrategy.cs
```diff
@@ -0,0 +1,152 @@
...
+                using var reg = new ManagementClass(scope, new ManagementPath("StdRegProv"), null);
+
+                foreach (var query in queries) {
+                    if (query.ValueNames == null) continue;
+
+                    var methodParams = reg.GetMethodParameters("EnumValues");
+                    methodParams["hDefKey"] = (UInt32)query.Hive;
+                    methodParams["sSubKeyName"] = query.KeyPath;
+
+                    using var outParams = reg.InvokeMethod("EnumValues", methodParams, null);
```
definitelynotagoblin, line 83: We don't need to worry about `EnumValues` raising an exception do we?  Would `InvokeMethod` even expose such a thing?

src/SharpHoundRPC/Registry/DotNetWmiRegistryStrategy.cs
```diff
@@ -0,0 +1,152 @@
...
+                RegistryValueKind.ExpandString => ("GetExpandedStringValue", "sValue"),
+                RegistryValueKind.Binary => ("GetBinaryValue", "uValue"),
+                RegistryValueKind.DWord => ("GetDWORDValue", "uValue"),
+                RegistryValueKind.MultiString => ("GetMultiStringValue", "sValue"),
+                RegistryValueKind.QWord => ("GetQWORDValue", "uValue"),
+                _ => throw new InvalidEnumArgumentException(
+                    $"Unsupported WMI registry value type. Type: {valueType}  Name: {valueName}")
+            };
+
+            using var outParams = reg.InvokeMethod(methodName, methodParams, null);
```
definitelynotagoblin, line 148: Same question about exception handling just to mark it here

src/SharpHoundRPC/Registry/StrategyExecutor.cs
```diff
@@ -0,0 +1,50 @@
...
+                } catch (Exception ex) {
+                    attempt.FailureReason = $"Collector failed: {ex.Message}.\nInner Exception: {ex.InnerException}";
+                }
+
+                attempts.Add(attempt);
+            }
+
+            return new StrategyExecutorResult<T> {
+                Results = null,
+                FailureAttempts = attempts
```
definitelynotagoblin, line 45: I think this will always have at least one attempt in attempts, even on first-time success.
I suspect the `attempts.Add(attempt)` just above is supposed to be in the `catch` block

definitelynotagoblin (approved):
But check my comment on StrategyExecutor

src/CommonLib/LdapUtils.cs
```diff
@@ -43,7 +44,7 @@ private readonly ConcurrentDictionary<string, string>
             new(StringComparer.OrdinalIgnoreCase);
 
         private readonly ILogger _log;
-        private readonly PortScanner _portScanner;
+        private readonly IPortScanner _portScanner;
```
ktstrader, line 47: There are also a couple of `new PortScanner()` that should also be IPortScanner below on lines 74, 79, and 81

ktstrader (approved):

src/CommonLib/LdapUtils.cs
rvazarkar: PortScanner is the concrete implementation of the interface, so that's intended

src/SharpHoundRPC/Registry/StrategyExecutor.cs
rvazarkar: The function returns on success so it won't add it

PR #206, opened 2025-05-28
[BED-5934] First pass to provide better SMB Signing query transparency in logging

definitelynotagoblin:
## Description
Provide trace logging for SMB info scanning to provide more visibility into query paths taken to get this data.
Add additional safety check to SMB2 response parsing to add further confidence that the packet returned is properly composed.

## Motivation and Context
BED-5934

## Types of changes

## Checklist:

### DIFF
--- a/src/CommonLib/SMB/SmbScanner.cs
+++ b/src/CommonLib/SMB/SmbScanner.cs
@@ -260,7 +260,7 @@ namespace SharpHoundCommonLib.SMB
             // Validate structure size of negotiate response
             var negotiateStructureSize = reader.ReadUInt16();
 
-            if (negotiateStructureSize != SMB2Constants.ExpectedNegotiateStructureSizeB)
+            if (negotiateStructureSize != SMB2Constants.ExpectedNegotiateStructureSizeA)
             {
                 _log.LogDebug($"Expected fixed-value SMB2 response structure size {SMB2Constants.ExpectedNegotiateStructureSizeB}, got {negotiateStructureSize}. Packet: {Convert.ToBase64String(responsePacket)}");
                 return (true, false);
--- a/src/CommonLib/SMB/SmbScanner.cs
+++ b/src/CommonLib/SMB/SmbScanner.cs
@@ -100,7 +100,8 @@ namespace SharpHoundCommonLib.SMB
                 // signingEnabled and signingRequired must both be True to confirm that SMB signing is required on this computer,
                 // so if we only collect one registry and it's False, we can conclude that SMB signing is not required
                 // but if it's True, we can't conclude either way and will return a Fail result
-                if ((requireRegistryValue == null && signingEnabled) || (enableRegistryValue == null && signingRequired))
+                // Note however that signingRequired implies signingEnabled
+                if (requireRegistryValue == null && signingEnabled)
                     return SharpHoundRPC.Result<SmbScanInfo>.Fail("Could not acquire enough registries to determine SMB Signing info");
 
                 return SharpHoundRPC.Result<SmbScanInfo>.Ok(new SmbScanInfo(host)
--- a/src/CommonLib/SMB/SmbScanner.cs
+++ b/src/CommonLib/SMB/SmbScanner.cs
@@ -7,9 +7,7 @@ using System.Threading.Tasks;
 using SharpHoundCommonLib.SMB.SMB1;
 using SharpHoundCommonLib.SMB.SMB2;
 using SharpHoundCommonLib.SMB.NetBIOS;
-using System.CodeDom;
 using Microsoft.Extensions.Logging;
-using SharpHoundRPC;
 using SharpHoundRPC.Registry;
 using Microsoft.Win32;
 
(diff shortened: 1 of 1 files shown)


src/CommonLib/SMB/SmbScanner.cs
```diff
@@ -256,7 +259,12 @@ internal bool CheckResponseRequiresSigning(byte[] responsePacket, bool parseSMB1
 
             // Validate structure size of negotiate response
             var negotiateStructureSize = reader.ReadUInt16();
-            
+
+            if (negotiateStructureSize != SMB2Constants.ExpectedNegotiateStructureSizeA)
```
definitelynotagoblin, line 266: Per https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/63abf97c-0d09-47e2-88d6-6bfa552949a5 it looks that StructureSize should always be valued 65, so we could use this as an additional sanity check on the response data if we wish.

That said, there's also a SMB2Constants.ExpectedNegotiateStructureSizeB with no code references and I can't determine its use or any documentation around it.

src/CommonLib/SMB/SmbScanner.cs
```diff
@@ -45,40 +43,50 @@ public SmbScanner(ILogger log)
...
-            return await TrySMBNegotiate(host, port, false);
+            // SMB1 failed or requires signing, so try an SMB2 negotiate in case it's vulnerable
+            _log.LogTrace($"Negotiating SMB2 request to host {host}");
+            var smb2result = await TrySMBNegotiate(host, port, false);
+
+            // Return SMB2 result, unless SMB1 is the only successful response
+            if (smb2result.IsSuccess || smb1result.IsFailed)
+                return smb2result;
+            else
+                return smb1result;
```
definitelynotagoblin: The above changes intend to accomplish these behaviors:
- If registry check fails or isn't confident, continue on to SMB Negotiate checks
- If _either_ SMB1 or SMB2 is enabled with signing not required, we return a signing not required response

src/CommonLib/SMB/SmbScanner.cs
```diff
@@ -45,40 +43,50 @@ public SmbScanner(ILogger log)
...
         }
 
         /// <summary>
         /// Determines if SMB signing is required using registry values.
         /// </summary>
         private SharpHoundRPC.Result<SmbScanInfo> CheckRegistrySigningRequired(string host)
         {
-            const string keyPath = @"SYSTEM\CurrentControlSet\Services\LanmanServer\Parameters";
+            const string keyPath = @"SYSTEM\CurrentControlSet\Services\LanManServer\Parameters";
```
definitelynotagoblin, line 80: This shouldn't do anything, but I was looking for holes 🤷

src/CommonLib/SMB/SmbScanner.cs
definitelynotagoblin: We'll back on SMBNegotiate if registry check fails, but won't touch the logic around SMB1-2 requests.

src/CommonLib/SMB/SmbScanner.cs
```diff
@@ -87,18 +86,22 @@ private SharpHoundRPC.Result<SmbScanInfo> CheckRegistrySigningRequired(string ho
...
-
                 if (enableRegistryValue != null)
-                {
                     signingEnabled = Convert.ToInt32(enableRegistryValue) != 0;
-                }
+
+                // signingEnabled and signingRequired must both be True to confirm that SMB signing is required on this computer,
+                // so if we only collect one registry and it's False, we can conclude that SMB signing is not required
+                // but if it's True, we can't conclude either way and will return a Fail result
+                if ((requireRegistryValue == null && signingEnabled) || (enableRegistryValue == null && signingRequired))
```
ktstrader: Is having `requireRegistryValue == null` and `enableRegistryValue == null` redundant since it already checks for them being null on line 89?

src/CommonLib/SMB/SmbScanner.cs
ktstrader: In the condition you reference `SMB2Constants.ExpectedNegotiateStructureSizeA` and the log is referencing `SMB2Constants.ExpectedNegotiateStructureSizeB`. Should they both be SizeA?

src/CommonLib/SMB/SmbScanner.cs
definitelynotagoblin: Sort of.  It's a little funky
We check if both are missing on line 89, in which case we know things went wrong, but here we're checking if specifically only one register is missing, and if so, do we have enough information to determine if signing is required given only the one known register?

This actually got simplified a little because we discovered that SigningRequired implies SigningEnabled.  So if SigningRequired is present, we know we have a valid configuration even if we can't find SigningEnabled, but not so for the opposite -- if SigningEnabled is true but SigningRequired is missing, we don't know enough to make any conclusions.

src/CommonLib/SMB/SmbScanner.cs
definitelynotagoblin: Good catch!  Thanks!

ktstrader (approved):

PR #213, opened 2025-06-09
[BED-5972] Add cancellation tokens to timeout logic

definitelynotagoblin:
## Description
`TimeoutAfter` will exit tasks that run over a time budget, but those tasks will continue to run in the background indefinitely.
Replacing with `ExecuteWithTimeout` to allow a cancellation token to be passed back into the executing task so that it may cooperatively end early if it's been cancelled.

### Changes
- Replace `TimeoutAfter` with `ExecuteWithTimeout`
- Add early exits to timeout tasks where appropriate
- Add timeout to `LookupPrincipalBySid` calls
- New tests in `CommonLibHelperTests.cs`
- `Dockerfile`: update .net version to target `CommonLibTest`
- Mark existing tests `[WindowsOnlyFact]` where appropriate

### How to read this PR
It looks more intimidating than it is.  Many of the changes simply add a `CancellationToken cancellationToken = default` arg to existing functions (due to `ISAMServer` changing).

Files to pay attention to:
- `Helpers.cs`: this is where the new ExecuteWithTimeout functions live
- `CommonLibHelperTests.cs`: new tests for ExecuteWithTimeout
- `LocalGroupProcessor.cs`: many TimeoutAfter replacements exist here, but at the bottom a new timeout is added to LookupPrincipalBySid

## Motivation and Context
https://specterops.atlassian.net/browse/BED-5972

## How Has This Been Tested?
Tested full collection on ESC 10

## Types of changes

## Checklist:


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM mono:6.12.0
 
 # Install .NET SDK
-ENV DOTNET_VERSION=7.0
+ENV DOTNET_VERSION=8.0
```
definitelynotagoblin, line 4: CommonLibTest targets 8.0

src/CommonLib/Helpers.cs
```diff
@@ -310,6 +321,102 @@ public static string DumpDirectoryObject(this IDirectoryObject directoryObject)
...
+        /// Returns a Fail result if a task runs longer than its budgeted time.
+        /// A cancellation token is passed to the executing function so it may exit cleanly if timeout is reached.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="timeout"></param>
+        /// <param name="func"></param>
+        /// <returns></returns>
+        public static async Task<Result<T>> ExecuteWithTimeout<T>(TimeSpan timeout, Func<CancellationToken, T> func) {
+            var cts = new CancellationTokenSource();
+            var task = Task.Run(() => func(cts.Token));
```
definitelynotagoblin: Should I pass the `cts.Token` into `Task.Run`?
Right now this task is guaranteed time in the thread pool, even if its timeout has already expired due to thread pool starvation shenanigans.  If I pass it the cancellation token, it may not run at all in those circumstances.

Making this change would speed things up and help alleviate stress on the thread pool when necessary, but at the cost of these tasks possibly giving up the ghost before they even try.

src/CommonLib/Helpers.cs
```diff
@@ -135,7 +139,8 @@ public static string DistinguishedNameToDomain(string distinguishedName) {
             int idx;
             if (distinguishedName.ToUpper().Contains("DELETED OBJECTS")) {
                 idx = distinguishedName.IndexOf("DC=", 3, StringComparison.Ordinal);
-            } else {
+            }
+            else {
```
definitelynotagoblin, line 143: Yeah my editor makes these format changes, and the dotnet formatter specifically doesn't allow me to not have new lines after closing curly braces :/
I'm working on it

src/CommonLib/Helpers.cs
```diff
@@ -310,6 +321,102 @@ public static string DumpDirectoryObject(this IDirectoryObject directoryObject)
...
+
+        /// <summary>
+        /// Returns a Fail result if a task runs longer than its budgeted time.
+        /// A cancellation token is passed to the executing function so it may exit cleanly if timeout is reached.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="timeout"></param>
+        /// <param name="func"></param>
+        /// <returns></returns>
+        public static async Task<Result<T>> ExecuteWithTimeout<T>(TimeSpan timeout, Func<CancellationToken, Task<T>> func) {
```
ktstrader, line 364: Would you be able the different use cases for this `ExecuteWithTimeout` and the above `ExecuteWithTimeout`? I gotta squint to see the slight difference between the 2 methods XD

definitelynotagoblin: I'm a fan that the two are effectively interchangeable.  Whether you're running an async or a sync function you just have to pass it into `ExecuteWithTimeout` and the compiler will choose the correct one for you, especially since their behaviors should be the same.

From the usage perspective this makes them "the same" method, but it does mean we're obfuscating some magic behind the scenes.  As long as they both work this shouldn't effect anybody, but if one breaks it could certainly be confusing to debug if you don't know which version to look at :/

definitelynotagoblin: I could give them different names, but the caller would have to figure out which one to use based on if they're passing in a synchronous or async function.

I guess they already have to do that for NetAPI and RPC methods tho.

definitelynotagoblin: Maybe I should add comments to the code above to communicate the distinction?

definitelynotagoblin: Try that out lemme know how that looks

ktstrader: Perfect! Thank you!

src/CommonLib/Helpers.cs
definitelynotagoblin: Elected to do this after all.
If a task isn't going to return anything the main thread is using anyway (at this time it's already returned a Timeout error), and cancellation has been requested, it may as well not run.

ktstrader (approved):

PR #217, opened 2025-06-17
[BED-5971] Let's add timeouts to blocking ops

definitelynotagoblin:
## Description
Adding timeout wrapper to many potentially hanging IO processes, especially where unmanaged code is involved.

## Motivation and Context
BED-5971
BED-5972

## How Has This Been Tested?
GOAD
ESC 10 lab

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Introduced a centralized timeout management utility to enforce operation timeouts and improve reliability across network, LDAP, and RPC calls.
  * Added asynchronous LDAP operations and modern async support for various network and directory operations.
  * Enhanced registry key and NTLM authentication handling with timeout and cancellation support.

* **Bug Fixes**
  * Improved handling of blocking external calls and potential timeouts, reducing the risk of application hangs during network or directory operations.

* **Refactor**
  * Unified and simplified timeout logic across components, replacing legacy or manual timeout mechanisms.
  * Updated method signatures and internal logic to use tuple returns and async patterns for better clarity and maintainability.
  * Improved variable naming and code style for better readability.

* **Tests**
  * Added comprehensive unit tests for the new timeout utility and updated existing tests to reflect changes in method signatures and timeout handling.
  * Removed obsolete tests related to deprecated timeout helpers.

* **Chores**
  * Updated workflow configuration to only trigger builds on pull requests, reducing unnecessary build executions.


src/CommonLib/ConnectionPoolManager.cs
```diff
@@ -24,33 +25,45 @@ public ConnectionPoolManager(LdapConfig config, ILogger log = null, IPortScanner
             _portScanner = scanner ?? new PortScanner();
         }
 
-        public IAsyncEnumerable<Result<string>> RangedRetrieval(string distinguishedName,
-            string attributeName, CancellationToken cancellationToken = new()) {
+        public async IAsyncEnumerable<Result<string>> RangedRetrieval(string distinguishedName,
+            string attributeName, [EnumeratorCancellation] CancellationToken cancellationToken = new()) {
             var domain = Helpers.DistinguishedNameToDomain(distinguishedName);
 
-            if (!GetPool(domain, out var pool)) {
```
definitelynotagoblin, line 31: While we could return to this pattern, it does make asyncing public methods harder in the future, so I'm happy to leave the change we're making

src/CommonLib/ConnectionPoolManager.cs
```diff
@@ -73,24 +86,24 @@ public void ReleaseConnection(LdapConnectionWrapper connectionWrapper, bool conn
             return (success, message);
         }
 
-        private bool GetPool(string identifier, out LdapConnectionPool pool) {
```
definitelynotagoblin, line 76: `out` vars aren't valid in async methods.  While this is not async due to the constituent methods within not being async (so we don't break our contract), if that ever changes it'll be easier to manage without the out vars.

You'll see a lot of that down the PR

src/CommonLib/Helpers.cs
```diff
@@ -296,8 +293,7 @@ public static RegistryResult GetRegistryKeyData(string target, string subkey, st
         }
 
         public static IRegistryKey OpenRemoteRegistry(string target) {
-            var key = new SHRegistryKey(RegistryHive.LocalMachine, target);
-            return key;
+            return SHRegistryKey.Connect(RegistryHive.LocalMachine, target).GetAwaiter().GetResult();
```
definitelynotagoblin, line 296: You'll see `GetAwaiter().GetResult()` sprinkled around.
I'd plumbed too greedily and too deep the async/await pattern, only noticing when all was said and done that I'd broken many public method contracts.

`GetAwaiter().GetResult()` should block and throw any exceptions raised just as any sychronous method, and allows me to undo the async/await at the highest level while leaving the rest intact.

That does mean we don't get any real benefit out of those async/awaits, but it should be easier to plumb async/await back into these public methods if we choose to change the public contract in the future

src/CommonLib/IRegistryKey.cs
definitelynotagoblin, line 1: SHRegistryKey is a wrapper around Win32.RegistryKey, but wasn't consistently used thru SHC.  I expanded it to fulfill the needs of other places RegistryKey was used, so they could be consolidated under SHRegistryKey

src/CommonLib/IRegistryKey.cs
```diff
@@ -1,34 +1,60 @@
...
 
-    public class SHRegistryKey : IRegistryKey
-    {
-        private RegistryKey _currentKey;
+    public class SHRegistryKey : IRegistryKey, IDisposable {
+        private readonly RegistryKey _currentKey;
 
-        public SHRegistryKey(RegistryHive hive, string machineName)
-        {
-            var remoteKey = RegistryKey.OpenRemoteBaseKey(hive, machineName);
```
definitelynotagoblin, line 16: Not a huge fan of blocking network operations happening in a constructor.  I moved this into a static async `Connect` method that returns a SHRegistryKey already connected.

src/CommonLib/LdapConnectionPool.cs
```diff
@@ -100,18 +100,21 @@ public async IAsyncEnumerable<LdapResult<IDirectoryObject>> Query(LdapQueryParam
 
                 try {
                     _log.LogTrace("Sending ldap request - {Info}", queryParameters.GetQueryInfo());
-                    response = (SearchResponse)connectionWrapper.Connection.SendRequest(searchRequest);
+                    response = await SendRequestWithTimeout(connectionWrapper.Connection, searchRequest);
```
definitelynotagoblin: I'm pretty excited to have timeouts around the bulk of the blocking Query, PagedQuery Ldap operations.

They're even async now.
That's nice

src/CommonLib/LdapConnectionPool.cs
```diff
@@ -949,5 +987,9 @@ private SearchRequest CreateSearchRequest(string distinguishedName, string ldapF
             searchRequest.Controls.Add(new SearchOptionsControl(SearchOption.DomainScope));
             return searchRequest;
         }
+
+        private async Task<SearchResponse> SendRequestWithTimeout(LdapConnection connection, SearchRequest request) {
+            return (SearchResponse)await connection.SendRequestAsync(request, TimeSpan.FromMinutes(2));
```
definitelynotagoblin: I don't precisely know what exception this throws when it times out (and I do hope it throws an exception when it times out), but there is a blanket `catch (Exception e)` block where it's called that should log the exception message if it pops up.

src/CommonLib/LdapUtils.cs
```diff
@@ -763,41 +811,45 @@ private static bool RequestNETBIOSNameFromComputer(string server, string domain,
...
                     }
 
                 var originEndpoint = new IPEndPoint(IPAddress.Any, 0);
+                cancellationToken.ThrowIfCancellationRequested();
+                // Blocking External Call
                 requestSocket.Bind(originEndpoint);
 
                 try {
+                    // Blocking External Call
                     requestSocket.SendTo(NameRequest, remoteEndpoint);
```
definitelynotagoblin, line 837: We might be able to replace the sync `SendTo` and `ReceiveFrom` functions with async alternatives, but the async pattern provided by SendToAsync, ReceiveFromAsync are older than the Task model, and I think older even than the Async Programming Model we wrapped in a Task model in `Extensions.cs`

And that means a lot of trouble, so for now we'll keep to the sync methods

src/CommonLib/LdapUtils.cs
```diff
@@ -1239,85 +1301,99 @@ private static string ComputeDisplayName(IDirectoryObject directoryObject, strin
                 case Label.Group:
                 case Label.Base:
                     if (!string.IsNullOrWhiteSpace(samAccountName)) {
-                        displayName = $"{samAccountName}@{domain}";    
-                    }else if (directoryObject.TryGetProperty(LDAPProperties.CanonicalName, out var canonicalName)) {
+                        displayName = $"{samAccountName}@{domain}";
+                    }
+                    else if (directoryObject.TryGetProperty(LDAPProperties.CanonicalName, out var canonicalName)) {
```
definitelynotagoblin, line 1306: Dangit but I think all of this below is just whitespace changes

src/CommonLib/Processors/ComputerSessionProcessor.cs
```diff
@@ -290,49 +290,33 @@ public async Task<SessionAPIResult> ReadUserSessionsRegistry(string computerName
...
-            RegistryKey key = null;
-
             try {
-                var task = OpenRegistryKey(computerName, RegistryHive.Users);
-
-                if (await Task.WhenAny(task, Task.Delay(10000)) != task) {
-                    _log.LogDebug("Hit timeout on registry enum on {Server}. Abandoning registry enum", computerName);
-                    ret.Collected = false;
-                    ret.FailureReason = "Timeout";
+                using (var key = await SHRegistryKey.Connect(RegistryHive.Users, computerName)) {
```
definitelynotagoblin, line 294: Here's where we replaced the Win32 RegistryKey with our own abstraction `SHRegistryKey`, so check for any regressions here please.

Doesn't look like it has a unit test, I'll look about adding one.

src/CommonLib/Processors/DomainTrustProcessor.cs
```diff
@@ -31,14 +31,11 @@ public async IAsyncEnumerable<DomainTrust> EnumerateDomainTrusts(string domain)
 
             // Attempt to get trust type
             var trustInfoList = new List<(string TargetName, System.DirectoryServices.ActiveDirectory.TrustType TrustType)>();
-            try
```
definitelynotagoblin, line 34: This try/catch was redundant, as GetDomain already hides away any Exceptions thrown.

src/CommonLib/SMB/SmbScanner.cs
definitelynotagoblin, line 1: The replacements in the new code with ExecuteWithTimeout should provide the same behaviors as the timeout logics before, but good to have extra eyes here.

src/CommonLib/IRegistryKey.cs
```diff
@@ -1,34 +1,60 @@
...
+        /// <param name="machineName"></param>
+        /// <returns></returns>
+        /// <exception cref="TimeoutException"></exception>
+        /// <exception cref="ArgumentException"></exception>
+        /// <exception cref="System.IO.IOException"></exception>
+        /// <exception cref="ArgumentNullException"></exception>
+        /// <exception cref="System.Security.SecurityException"></exception>
+        /// <exception cref="UnauthorizedAccessException"></exception>
+        public static async Task<SHRegistryKey> Connect(RegistryHive hive, string machineName) {
+            var remoteKey = await Timeout.ExecuteWithTimeout(TimeSpan.FromSeconds(10), (_) => RegistryKey.OpenRemoteBaseKey(hive, machineName));
```
mykeelium, line 38: Should this function also set the _currentKey value of the class on success? Or is that not how this is used?

definitelynotagoblin: The `return new SHRegistryKey(remoteKey.Value)` should return a SHRegistryKey with the _currentKey value set per
``` cs
private SHRegistryKey(RegistryKey registryKey) {
    _currentKey = registryKey;
}
```

definitelynotagoblin: It's a little funky because it's a static method that returns an instance of the very class it's defined in.  But I didn't like the public constructor doing network work (cause who expects to put a try/catch around a constructor anyway? or that a constructor might block?), so I choose this route 🤷

src/CommonLib/Helpers.cs
definitelynotagoblin: Something I realized well in retrospect: I can create async methods and use those internally, exposing the sync methods just as simple `GetAwaiter().GetResult()` wrappers around them.

Definitely will want to do that, but that'll be an after prerelease 1 sorta job.

src/CommonLib/Ntlm/HttpNtlmAuthenticationService.cs
```diff
@@ -122,26 +139,35 @@ private async Task AuthWithBadChannelBindingsAsync(Uri url, string authScheme) {
...
-        var credentialCache = new CredentialCache();
-        credentialCache.Add(url, authScheme, CredentialCache.DefaultNetworkCredentials);
+        var credentialCache = new CredentialCache {
+            { url, authScheme, CredentialCache.DefaultNetworkCredentials }
+        };
 
         handler.Credentials = credentialCache;
         handler.PreAuthenticate = true;
 
         using var client = new HttpClient(handler);
```
mykeelium, line 154: I am not sure how relevant this is not in web development, but I believe this strategy might be susceptible to socket exhaustion. I have not seen it done with a supplied HttpClientHandler however, and that might rectify the issue. I need to look at this more deeply to understand if it needs, and if so, how, to mitigate.

ktstrader (approved):

src/CommonLib/IRegistryKey.cs
mykeelium: I see, yeah I totally missed the static for some reason. Thank you.

src/CommonLib/Processors/DomainTrustProcessor.cs
mykeelium: It looks like GetAllTrustRelationships can throw exceptions. It looks like that is what the log is referencing. 

https://learn.microsoft.com/en-us/dotnet/api/system.directoryservices.activedirectory.domain.getalltrustrelationships?view=windowsdesktop-9.0

src/CommonLib/Timeout.cs
```diff
@@ -0,0 +1,179 @@
...
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <param name="timeout"></param>
+    /// <param name="func"></param>
+    /// <returns></returns>
+    public static async Task<Result<T>> ExecuteWithTimeout<T>(TimeSpan timeout, Func<CancellationToken, T> func, CancellationToken parentToken = default) {
+        // cts will cancel its token if the parentToken is cancelled
+        // a default parentToken will never cancel so should noop this CreateLinkedTokenSource
+        var cts = CancellationTokenSource.CreateLinkedTokenSource(parentToken);
+        var task = Task.Factory.StartNew(() => func(cts.Token), cts.Token, TaskCreationOptions.None, TaskScheduler.Current);
```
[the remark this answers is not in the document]
definitelynotagoblin: I think we want to make it easier for potential custom schedulers to be specified in the future for synchronous operations.  I'm inclined to leave it as is, but I suppose either way we need to be aware of this specific detail of implementation. 🤔

src/CommonLib/Timeout.cs
```diff
@@ -0,0 +1,179 @@
...
+                return Result<T>.Ok(await task);
+            }
+            catch (OperationCanceledException) { }
+        }
+
+        if (parentToken.IsCancellationRequested)
+            return Result<T>.Fail("Cancellation requested");
+        else
+            return Result<T>.Fail("Timeout");
+    }
```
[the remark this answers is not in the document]
definitelynotagoblin: In initial tests this was causing issues of the cancellation token being disposed before a task could check it.
In retrospect however this should be a non-issue - any task checking a disposed cancellation token has already been "orphaned" by the timeout handler.  I'll reintroduce this 👍

src/CommonLib/LdapConnectionPool.cs
```diff
@@ -645,7 +673,7 @@ private bool CallDsGetDcName(string domainName, out NetAPIStructs.DomainControll
 
         public (bool Success, LdapConnectionWrapper connectionWrapper, string Message)
             GetConnectionForSpecificServerAsync(string server, bool globalCatalog) {
-            return CreateNewConnectionForServer(server, globalCatalog);
+            return CreateNewConnectionForServer(server, globalCatalog).GetAwaiter().GetResult();
         }
```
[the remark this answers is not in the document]
definitelynotagoblin: I only wish we could plumb the async/await pattern up here 😞 
But until we decide to update the public contract of this library, we're stuck with the synchronous method.
I'd plumbed everything to these public methods with async/await, with the intent that one day we could update this contract.

Hmmm.  What I can do is create a private async overload this library uses internally, and expose the public synchronous method as just a blocking call on the async one.  I think I'll go do that.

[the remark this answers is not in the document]
definitelynotagoblin: GetConnectionForSpecificServerAsync isn't even async 😭

src/CommonLib/SMB/SmbScanner.cs
```diff
@@ -126,38 +113,32 @@ private SharpHoundRPC.Result<SmbScanInfo> CheckRegistrySigningRequired(string ho
...
+                else {
                     var negotiateRequest = new SMB2NegotiateRequest(host);
                     negoReqBytes = negotiateRequest.ToBytes();
                 }
 
-                var negoRespBytes = await SendAndReceiveData(host, port, TimeoutMs, negoReqBytes);
-
+                var negoResp = await Timeout.ExecuteWithTimeout(TimeSpan.FromMilliseconds(TimeoutMs), (timeoutToken) => SendAndReceiveData(host, port, negoReqBytes, timeoutToken));
+                if (!negoResp.IsSuccess)
+                    throw new OperationCanceledException("Connection attempt timed out");
```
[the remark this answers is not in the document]
definitelynotagoblin: In this case I believe negoResp.IsSuccess will only be false when a timeout occurs, and the throw should emulate behavior before this changeset.  Is that mistaken?

src/CommonLib/Processors/DomainTrustProcessor.cs
definitelynotagoblin: I think you're right this comment is out of date.
Good catch, reverting.

src/CommonLib/Ntlm/HttpNtlmAuthenticationService.cs
definitelynotagoblin: @coderabbitai, my understanding of sockets is pretty skeletal.  You have any opinions, or words to help me understand?

mykeelium (approved):

Issue #203, opened 2025-05-04
Performance degradation when SharpHound attempts to resolve unresolvable SIDs


NexKosh:
When running SharpHound from a non-domain-joined Windows machine using supplied credentials, if target objects in the domain contain unresolvable SIDs (e.g., due to stale entries from broken two-way trusts), the tool repeatedly attempts to resolve them.

- Environment

1. SharpHound Version: 2.6.5
2. SharpHoundCommon Version: 4.2.6
3. Command : SharpHound -c DCOnly --ldapusername domainuser@ludus.domain --ldappassword ******** -d ludus.domain -v 1 --skipportcheck --disablecertverification --disablesigning --domaincontroller ludus.domain

- Conditions

1. The machine executing SharpHound is not joined to the domain
2. Some AD objects contain SIDs from a previously trusted domain (e.g., due to removed two-way trust)
3. unresolvable SID (from a previously trusted domain) was present on many AD objects.
4. Total object count is large (e.g., 500,000+ objects)
- Screenshot

1. non-domain-joined ![Image](https://github.com/user-attachments/assets/124d552f-4d72-4375-b9fe-99244da0a757)

2. domain-joined ![Image](https://github.com/user-attachments/assets/556e3552-3bbd-4129-ba73-71a01ccea198)

- Observed Behavior
1. SharpHound attempts to resolve unknown or external SIDs via DirectoryContext using GetDomain(...)
2. Each failed resolution logs:
`System.DirectoryServices.ActiveDirectory.ActiveDirectoryOperationException: Current security context is not associated with an Active Directory domain or forest.
`
- Comparison Results

Scenario | Cache | Execution Time
-- | -- | --
Domain-joined host | ❌ / ✅ | 13 mins
Non-domain host, same creds | ✅ | 30 mins
Non-domain host, same creds | ❌ | 90 mins

Is it possible to add unresolvable SID cache to avoid repeated requests or disable try get unresolvable SID?

ktstrader:
Hi @NexKosh! Would you be able to provide steps to reproduce this issue for testing purposes?

NexKosh:
Hi @ktstrader 👋

To accurately simulate SharpHound performance degradation caused by unresolved or stale ACL SIDs (typically from removed trusts or deleted users), I suggest the following setup:

✅ Recommended Procedure

1. **Establish a Two-Way Trust Between Domains**

[...]

PR #219, opened 2025-06-27
fix: BED-6026 adding the domainSid to _unresolvablePrincipals cache to prevent repeated resolving

ktstrader:
## Description

SH tries to repeatedly resolve unresolvable SIDs from a non-domain-joined Windows machine.

## Motivation and Context

https://specterops.atlassian.net/browse/BED-6026
https://github.com/SpecterOps/SharpHoundCommon/issues/203

## Types of changes

## Checklist:

### DIFF
--- a/src/CommonLib/LdapUtils.cs
+++ b/src/CommonLib/LdapUtils.cs
@@ -383,6 +383,8 @@ namespace SharpHoundCommonLib {
 
         private async Task<(bool Success, string DomainName)> ConvertDomainSidToDomainNameFromLdap(string domainSid) {
             if (!GetDomain(out var domain) || domain?.Name == null) {
+                // fix: BED-6026 adding the domainSid to _unresolvablePrincipals cache to prevent repeated resolving
+                _unresolvablePrincipals.Add(domainSid);
                 return (false, string.Empty);
             }
 


src/CommonLib/LdapUtils.cs
```diff
@@ -383,6 +383,8 @@ public IAsyncEnumerable<LdapResult<IDirectoryObject>> PagedQuery(LdapQueryParame
 
         private async Task<(bool Success, string DomainName)> ConvertDomainSidToDomainNameFromLdap(string domainSid) {
             if (!GetDomain(out var domain) || domain?.Name == null) {
+                // fix: BED-6026 adding the domainSid to _unresolvablePrincipals cache to prevent repeated resolving
+                _unresolvablePrincipals.Add(domainSid);
```
ktstrader, line 390: GetDomain returns false if the external call returns an exception. So I figured we can add the SID to _unresolvablePrincipals when it checks the domain and it comes back blank. This can avoid any other attempts to try and resolve that SID and lower the execution time that the user had stated they were experiencing.

ktstrader: I have not been able to test this. Mini GOAD has let me down

PR #222, opened 2025-07-09
[BED-5972] Log timeout-wrapped execution time stats, add adaptive timeout logic

definitelynotagoblin:
## Description
A first step towards execution time analytics.
Adding self-configuring timeouts to existing timeout wrappers.

## Motivation and Context
https://specterops.atlassian.net/browse/BED-5972
https://specterops.atlassian.net/browse/BED-6198

## How Has This Been Tested?
Unit tests, local lab

## Screenshots (if appropriate):
image

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Introduced adaptive timeout handling across various operations, replacing fixed timeout values with dynamic, self-adjusting timeouts based on recent execution times.
  * Added new adaptive timeout and execution time sampling utilities for improved reliability and performance.

* **Improvements**
  * Updated multiple processor and utility classes to use adaptive timeouts, enhancing robustness against network variability and operational delays.
  * Simplified method signatures by removing explicit timeout parameters where adaptive timeouts are now applied.
  * Enhanced cancellation support for timeout operations.
  * Centralized and standardized timeout management in network, RPC, and LDAP calls.
  * Improved LDAP request handling with adaptive timeouts and retry logic for timeouts.

* **Bug Fixes**
  * Improved timeout management to better handle sudden spikes and prevent skewed timeout calculations.

* **Tests**
  * Added comprehensive unit tests for adaptive timeout logic.
  * Updated or commented out obsolete timeout-related tests to reflect new adaptive timeout mechanisms.

### DIFF
--- a/src/CommonLib/AdaptiveTimeout.cs
+++ b/src/CommonLib/AdaptiveTimeout.cs
@@ -181,10 +181,16 @@ public sealed class AdaptiveTimeout : IDisposable {
         if (!_useAdaptiveTimeout || _sampler.Count < _minSamplesForAdaptiveTimeout)
             return _maxTimeout;
 
-        var stdDev = _sampler.StandardDeviation();
-        var adaptiveTimeoutMs = _sampler.Average() + (stdDev * StdDevMultiplier);
-        var cappedTimeoutMS = Math.Min(adaptiveTimeoutMs, _maxTimeout.TotalMilliseconds);
-        return TimeSpan.FromMilliseconds(cappedTimeoutMS);
+        try {
+            var stdDev = _sampler.StandardDeviation();
+            var adaptiveTimeoutMs = _sampler.Average() + (stdDev * StdDevMultiplier);
+            var cappedTimeoutMS = Math.Min(adaptiveTimeoutMs, _maxTimeout.TotalMilliseconds);
+            return TimeSpan.FromMilliseconds(cappedTimeoutMS);
+        }
+        catch (Exception ex) {
+            _log.LogError(ex, "Error calculating adaptive timeout, defaulting to max timeout.");
+            return _maxTimeout;
+        }
     }
 
     // AdaptiveTimeout will not respond well to rapid spikes in execution time
--- a/src/CommonLib/AdaptiveTimeout.cs
+++ b/src/CommonLib/AdaptiveTimeout.cs
@@ -170,15 +170,17 @@ public sealed class AdaptiveTimeout : IDisposable {
         _sampler.Dispose();
     }
 
-    // Within 5 standard deviations will have a conservative lower bound of catching 98% of executions (1 - 1/(k^2 / 2)),
+    // Within 5 standard deviations will have a conservative lower bound of catching 96% of executions (1 - 1/5^2),
     // regardless of sample shape
     // so long as those samples are independent and identically distributed
     // (and if they're not, our TimeSpikeSafetyValve should provide us with some adaptability)
-    // But we'll cap at configured timeout
+    // But the effective collection rate is probably closer to 98+%
+    // (in part because we don't need to filter out "too fast" outliers)
+    // But we'll cap at configured maximum timeout
     // https://modelassist.epixanalytics.com/space/EA/26574957/Tchebysheffs+Rule
     // https://en.wikipedia.org/wiki/Independent_and_identically_distributed_random_variables
     public TimeSpan GetAdaptiveTimeout() {
-        if (!_useAdaptiveTimeout || _sampler.Count < _minSamplesForAdaptiveTimeout)
+        if (!UseAdaptiveTimeout())
             return _maxTimeout;
 
         try {
@@ -206,8 +208,9 @@ public sealed class AdaptiveTimeout : IDisposable {
         else
             _clearSamplesDecay += TimeSpikePenalty;
 
+
         if (_clearSamplesDecay >= ClearSamplesThreshold) {
-            if (_useAdaptiveTimeout && _sampler.Count >= _minSamplesForAdaptiveTimeout) {
+            if (UseAdaptiveTimeout()) {
                 ClearSamples();
                 _log.LogTrace("Time spike safety valve event at timeout {CurrentTimeout}.", GetAdaptiveTimeout());
             }
@@ -215,5 +218,9 @@ public sealed class AdaptiveTimeout : IDisposable {
                 _log.LogWarning("This call is frequently running over the maximum allowed timeout of {MaxTimeout}.", _maxTimeout);
             }
         }
+    }
+
+    private bool UseAdaptiveTimeout() {
+        return _useAdaptiveTimeout && _sampler.Count >= _minSamplesForAdaptiveTimeout;
     }
 }
--- a/src/CommonLib/AdaptiveTimeout.cs
+++ b/src/CommonLib/AdaptiveTimeout.cs
@@ -18,7 +18,7 @@ public sealed class AdaptiveTimeout : IDisposable {
     private const int ClearSamplesThreshold = 5;
     private const int StdDevMultiplier = 5;
 
-    public AdaptiveTimeout(TimeSpan maxTimeout, ILogger log, int sampleCount, int logFrequency, int minSamplesForAdaptiveTimeout, bool useAdaptiveTimeout = true) {
+    public AdaptiveTimeout(TimeSpan maxTimeout, ILogger log, int sampleCount = 100, int logFrequency = 1000, int minSamplesForAdaptiveTimeout = 30, bool useAdaptiveTimeout = true) {
         if (maxTimeout <= TimeSpan.Zero)
             throw new ArgumentException("maxTimeout must be positive", nameof(maxTimeout));
         if (sampleCount <= 0)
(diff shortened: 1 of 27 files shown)


src/CommonLib/ExecutionTimeSampler.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    public double StandardDeviation() {
+        double average = _samples.Average();
+        double sumOfSquaresOfDifferences = _samples.Select(val => (val - average) * (val - average)).Sum();
+        double stddiv = Math.Sqrt(sumOfSquaresOfDifferences / _samples.Count);
+
+        return stddiv;
+    }
+
+    public double Average() => _samples.Average();
```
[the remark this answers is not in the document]
definitelynotagoblin: Nah, our values don't need to be precise for analysis, not worried about thread safety here.

src/CommonLib/AdaptiveTimeout.cs
```diff
@@ -0,0 +1,208 @@
...
+        var result = await Timeout.ExecuteRPCWithTimeout(GetAdaptiveTimeout(), (timeoutToken) => _sampler.SampleExecutionTime(() => func(timeoutToken)), parentToken);
+        TimeSpikeSafetyValve(result.IsSuccess);
+        return result;
+    }
+
+    public void Dispose() {
+        _sampler.Dispose();
+    }
+
+    // Within 5 standard deviations will have a conservative lower bound of catching 98% of executions (1 - 1/(k^2 / 2)),
```
mykeelium: I do not see this " / 2" in the Tchebysheffs rule. Is there something else here going on?

src/CommonLib/AdaptiveTimeout.cs
```diff
@@ -0,0 +1,208 @@
...
+    // regardless of sample shape
+    // so long as those samples are independent and identically distributed
+    // (and if they're not, our TimeSpikeSafetyValve should provide us with some adaptability)
+    // But we'll cap at configured timeout
+    // https://modelassist.epixanalytics.com/space/EA/26574957/Tchebysheffs+Rule
+    // https://en.wikipedia.org/wiki/Independent_and_identically_distributed_random_variables
+    public TimeSpan GetAdaptiveTimeout() {
+        if (!_useAdaptiveTimeout || _sampler.Count < _minSamplesForAdaptiveTimeout)
```
mykeelium: There is a try catch for the log ExecutionTimeSampler.Log for calling Average and StandardDeviation. Should this method also try catch? Maybe respond with maxTimeout or should it throw?

src/CommonLib/Processors/DCLdapProcessor.cs
```diff
@@ -22,43 +22,41 @@ public class LdapAuthOptions {
...
     public delegate Task ComputerStatusDelegate(CSVComputerStatus status);
 
     private readonly string SEC_E_UNSUPPORTED_FUNCTION = "80090302";
     private readonly string SEC_E_BAD_BINDINGS = "80090346";
 
 
-    public DCLdapProcessor(int portScanTimeout, string dcHostname, ILogger log = null) {
+    public DCLdapProcessor(int ldapConnectionTimeoutMs, string dcHostname, ILogger log = null) {
         _log = log ?? Logging.LogProvider.CreateLogger("DCLdapProcessor");
         _scanner = new PortScanner();
```
mykeelium: Should this portScanner be created with the supplied timeout since it was removed from the function calls on lines 118 and 123 on this file?

src/CommonLib/AdaptiveTimeout.cs
mykeelium: Seeing how this could be used in a log such as in PortScanner, if there is a possibility for this to throw it might be best to use a try catch.

mykeelium (commented):
This looks like an awesome change and extremely well done for how large it is! This will be great to get in. A few comment and questions. Since some interfaces changed should this also be a minor version bump?

src/CommonLib/AdaptiveTimeout.cs
definitelynotagoblin: Ah, that's just because Tchebysheffs specifies bounds for a random data point to be within X standard deviations of the mean, but we don't actually care about processes that are "running too fast" - we're including all values less than the mean, so if we assume a symmetrical time plot, we're not trimming off the bottom outliers.  So while / 2 isn't precise, I think it's a fair approximation.

src/CommonLib/AdaptiveTimeout.cs
definitelynotagoblin: Good point, I'll add that :+1:

src/CommonLib/Processors/DCLdapProcessor.cs
definitelynotagoblin: Oh good catch, thanks!

definitelynotagoblin:
Honestly I've been on the fence about that too.
I may have been overzealous in removing the now-unused timeout params from the public functions.
I'm inclined to say yes just cover it under a minor version bump,
But I'll have to confirm above if we're okay with that.

src/CommonLib/AdaptiveTimeout.cs
definitelynotagoblin: Average() should never throw an exception so long as `_sampler.Count >= _minSamplesForAdaptiveTimeout`, but better safe than sorry.

src/CommonLib/ExecutionTimeSampler.cs
```diff
@@ -0,0 +1,104 @@
...
+        Log(flush: true);
+        _samples = new ConcurrentQueue<double>();
+    }
+
+    public double StandardDeviation() {
+        double average = _samples.Average();
+        double sumOfSquaresOfDifferences = _samples.Select(val => (val - average) * (val - average)).Sum();
+        double stddiv = Math.Sqrt(sumOfSquaresOfDifferences / _samples.Count);
+
+        return stddiv;
```
definitelynotagoblin, line 40: This is a very common operation, and kinda expensive, but I would hazard to say that we are far more network bottlenecked than CPU bottlenecked, so I'm not worried about it currently.

src/CommonLib/AdaptiveTimeout.cs
```diff
@@ -0,0 +1,219 @@
...
+        else
+            _clearSamplesDecay += TimeSpikePenalty;
+
+        if (_clearSamplesDecay >= ClearSamplesThreshold) {
+            if (_useAdaptiveTimeout && _sampler.Count >= _minSamplesForAdaptiveTimeout) {
+                ClearSamples();
+                _log.LogTrace("Time spike safety valve event at timeout {CurrentTimeout}.", GetAdaptiveTimeout());
+            }
+            else {
+                _log.LogWarning("This call is frequently running over the maximum allowed timeout of {MaxTimeout}.", _maxTimeout);
```
definitelynotagoblin, line 218: This warning may get noisy, but then we also shouldn't be seeing it unless Something Is Wrong ™️

src/CommonLib/AdaptiveTimeout.cs
```diff
@@ -0,0 +1,219 @@
...
+        var result = await Timeout.ExecuteRPCWithTimeout(GetAdaptiveTimeout(), (timeoutToken) => _sampler.SampleExecutionTime(() => func(timeoutToken)), parentToken);
+        TimeSpikeSafetyValve(result.IsSuccess);
+        return result;
+    }
+
+    public void Dispose() {
+        _sampler.Dispose();
+    }
+
+    // Within 5 standard deviations will have a conservative lower bound of catching 98% of executions (1 - 1/(k^2 / 2)),
```
definitelynotagoblin: Actually I think technically this is a lower bound of 96%, but that's very conservative and assumes a very non-normal distribution.
I'm specifying `1/(k^2 / 2)` because Tchbysheffs' measures how many samples are _within_ so much distance of the mean, but we're not concerned with samples running "too fast" so if we assume this plot is somewhat symmetrical, I think assuming about as many samples are too fast as too slow is appropriate.

Hence 4% of samples (`1/5^2`) are outliers on one end or the other, approximately 2% to each side.

definitelynotagoblin: Cleared this up :+1:

src/CommonLib/Processors/LocalGroupProcessor.cs
```diff
@@ -16,22 +16,33 @@ public class LocalGroupProcessor
...
-        {
+        private readonly AdaptiveTimeout _getMachineSidAdaptiveTimeout;
+        private readonly AdaptiveTimeout _openSamServerAdaptiveTimeout;
+        private readonly AdaptiveTimeout _getDomainsAdaptiveTimeout;
+        private readonly AdaptiveTimeout _openDomainAdaptiveTimeout;
+        private readonly AdaptiveTimeout _getAliasesAdaptiveTimeout;
+        private readonly AdaptiveTimeout _openAliasAdaptiveTimeout;
+        private readonly AdaptiveTimeout _getMembersAdaptiveTimeout;
+        private readonly AdaptiveTimeout _lookupPrincipalBySidAdaptiveTimeout;
+
```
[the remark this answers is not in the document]
definitelynotagoblin: While this is technically correct, the Dispose here is just to flush analytics logs for the ExecutionTimeSampler, and I'm not too worried if we're missing some of the tail end of these once LocalGroupProcessor is cleaned up - at least not enough to add more complexity to this class.
The IDisposable interface is used more as a courtesy than as a safety in this case.

mykeelium (approved):

PR #242, opened 2025-08-27
[BED-6389] Add retry logic to GetDomains to make more resilient to network hiccups

definitelynotagoblin:
## Motivation and Context
When the static GetDomain call fails in CreateNewConnection, that domain is immediately added to the excluded domains list for the remainder of the job.  Maybe this should be a little more forgiving.
https://specterops.atlassian.net/browse/BED-6389

## Types of changes

## Checklist:

## Summary by CodeRabbit

- New Features
  - Configurable retry delays with decorrelated jitter, including support for retrying functions that return values.
  - Thread-safe random utilities for generating doubles and longs.

- Improvements
  - Centralized LDAP domain exclusion with early-exit checks and automatic exclusion after failures.
  - Safer delay handling with bounds to prevent overflow.

- Tests
  - Added comprehensive backoff timing tests.
  - Updated LDAP connection pool tests to reflect new exclusion behavior.

### DIFF
--- a/src/CommonLib/Helpers.cs
+++ b/src/CommonLib/Helpers.cs
@@ -340,6 +340,23 @@ namespace SharpHoundCommonLib {
                 }
             } while (!success && attempt < retryCount);
         }
+
+        public static U RetryOnException<T, U>(Func<U> action, int retryCount, ILogger logger = null) where T : Exception {
+            int attempt = 0;
+            do {
+                try {
+                    return action();
+                }
+                catch (T e) {
+                    attempt++;
+                    logger?.LogDebug(e, "Exception caught, retrying attempt {Attempt}", attempt);
+                    if (attempt >= retryCount)
+                        throw;
+                }
+            } while (attempt < retryCount);
+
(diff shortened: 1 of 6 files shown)


src/CommonLib/Helpers.cs
```diff
@@ -318,15 +320,28 @@ public static string DumpDirectoryObject(this IDirectoryObject directoryObject)
             return builder.ToString();
         }
 
+        public static TimeSpan BackoffWithDecorrelatedJitter(int attempt, TimeSpan baseDelay, TimeSpan maxDelay) {
+            // Decorrelated Jitter Backoff - see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+            var temp = Math.Min(maxDelay.Ticks, baseDelay.Ticks * Math.Pow(2, attempt));
```
definitelynotagoblin: Nit: Math.Pow is inherently very slow since it always works in floating-point space.  Would recommend `attempt * attempt` here

mykeelium: Thank you for this. Updated this to use long rather than Math.Pow. Please take a look at the new functionality. Thank you

PR #258, opened 2025-11-06
BED-5558 - SharpHound warn on inability to collect necessary properties

mykeelium:
## Description

This stops SharpHound from setting property values on users when we were unable to collect User Account Control. A Log Warning has been added to explicitly state this. 

## Motivation and Context

BED-5558

## How Has This Been Tested?

This was tested by creating a build, running collection, and ingesting the collection. In addition the Unit test passed locally. 

## Types of changes

## Checklist:

-   [x] Documentation updates are needed, and have been made accordingly.

## Summary by CodeRabbit

* **New Features**
  * Added optional runtime logging support to improve diagnostic output.

* **Bug Fixes**
  * Improved handling and validation of account delegation data, skipping invalid entries and emitting clearer warnings when required flags are missing.
  * Strengthened SID validation for delegated hosts to reduce false positives.

* **Tests**
  * Updated unit tests to match the tightened property validation and delegation behavior.

### DIFF
--- a/src/CommonLib/Processors/LdapPropertyProcessor.cs
+++ b/src/CommonLib/Processors/LdapPropertyProcessor.cs
@@ -265,11 +265,18 @@ namespace SharpHoundCommonLib.Processors {
                             continue;
 
                         var resolvedHost = await _utils.ResolveHostToSid(d, domain);
-                        if (resolvedHost.Success && resolvedHost.SecurityIdentifier.Contains("S-1"))
-                            comps.Add(new TypedPrincipal {
-                                ObjectIdentifier = resolvedHost.SecurityIdentifier,
-                                ObjectType = Label.Computer
-                            });
+                        if (!resolvedHost.Success || !resolvedHost.SecurityIdentifier.Contains("S-1")) continue;
+                        await SendComputerStatus(new CSVComputerStatus {
+                            Status = CSVComputerStatus.StatusSuccess,
+                            Task = nameof(ReadUserProperties),
+                            ComputerName = Helpers.StripServicePrincipalName(d).ToUpper().TrimEnd('$'),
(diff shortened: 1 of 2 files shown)


src/CommonLib/Processors/LdapPropertyProcessor.cs
```diff
@@ -233,55 +235,57 @@ public async Task<UserProperties> ReadUserProperties(IDirectoryObject entry, str
...
-                        continue;
+                    foreach (var d in delegates) {
+                        if (d == null)
+                            continue;
 
-                    var resolvedHost = await _utils.ResolveHostToSid(d, domain);
-                    if (resolvedHost.Success && resolvedHost.SecurityIdentifier.Contains("S-1"))
-                    {
+                        var resolvedHost = await _utils.ResolveHostToSid(d, domain);
+                        if (!resolvedHost.Success || !resolvedHost.SecurityIdentifier.Contains("S-1")) continue;
```
JimSycurity: What if my computer is named "SALES-100", but the SID cannot be resolved, will that be handled properly?

ie, is the Contains() method the best option here?  Would StartsWith() be a better option and expand the comparator to "S-1-5-" to represent the full NT Authority namespace in the SID structure that all computers should belong to?

SID namespace reference: https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers

mykeelium: If I understand it, if the SID cannot be resolved, this should handle that, as it checking for success. That being said, I am not entirely sure why it was a contains, and not a starts with. I updated this and the other place in the file that had this. I also updated it so "S-1-5-" as you recommended.

definitelynotagoblin (approved):

PR #261, opened 2025-11-14
fix: add adaptive timeouts to missing portions of enumeration

rvazarkar:
## Description

The check for IsWebClientRunning was not wrapped in a timeout. Additionally, our registry strategy executor was not wrapped in timeouts either. Both of these are used during SHE local group collection and could potentially lead to major hangs.

## Motivation and Context
https://specterops.atlassian.net/browse/BED-6829

## How Has This Been Tested?

Unable to reproduce, so adding these as a precaution.

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal control flow to reduce branching and improve readability.

* **Bug Fixes**
  * Added adaptive timeout protection (2-minute max) to registry and web-client operations to prevent hangs and return clear timeout failures.

### DIFF
--- a/src/CommonLib/Processors/WebClientServiceProcessor.cs
+++ b/src/CommonLib/Processors/WebClientServiceProcessor.cs
@@ -88,6 +88,7 @@ namespace SharpHoundCommonLib.Processors {
                 }
             });
 
+            //TODO: Not a big fan of nested result objects. We should look at this later to see if we can simplify the interface
             if (result.IsSuccess) {
                 return result.Value;
             }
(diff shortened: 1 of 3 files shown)


src/CommonLib/Processors/WebClientServiceProcessor.cs
```diff
@@ -78,15 +79,20 @@ public async Task<APIResult<bool>> IsWebClientRunning(string computerName) {
...
                 try {
                     var exists = TestPathExists(pipePath);
-
                     return APIResult<bool>.Success(exists);
                 } catch (Exception ex) {
                     return APIResult<bool>.Failure(ex.ToString());
                 }
             });
+
+            if (result.IsSuccess) {
```
mykeelium, line 92: nit (non-blocking): The stacking of results here is a bit confusing. I am not sure if we should notate that, and in the future possibly find a way to flatten this functionality.

mykeelium (approved):

PR #279, opened 2026-03-11
chore: Improve build workflow BED-7214

lrfalslev:
## Description

PR check runs test suite.
Merge publishes a `[v].[v].[v]-dev.[date]` nuget package.
Normalize pr template across SH repos.

## Motivation and Context

[BED-7214](https://specterops.atlassian.net/browse/BED-7214)

## How Has This Been Tested?

Validated workflows are passing.
Triggered publish dev package workflow and confirmed -dev packages are published and pruned.

## Screenshots (if appropriate):

image
image
image

## Types of changes

## Checklist:

## Summary by CodeRabbit

* **Documentation**
  * Improved pull request template with clearer guidance, structured sections, and an actionable checklist for contributors.

* **Chores**
  * Added a "Build and Test" CI workflow to run build and tests on pull requests.
  * Added a "Publish Dev Package" CI workflow to pack and prepare dev packages (publish step dry-run).
  * Introduced a minor public API change; tests were updated accordingly.

[BED-7214]: https://specterops.atlassian.net/browse/BED-7214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/publish-dev-package.yml
+++ b/.github/workflows/publish-dev-package.yml
@@ -45,4 +45,4 @@ jobs:
           dotnet tool install -g sleet
           sleet push ./pkgs
           sleet retention prune --package SharpHoundCommon --package SharpHoundRPC `
-            --stable 999 --prerelease 5 --release-labels 1
+            --release-labels 1 --dry-run
--- a/.github/workflows/publish-dev-package.yml
+++ b/.github/workflows/publish-dev-package.yml
@@ -45,4 +45,4 @@ jobs:
           dotnet tool install -g sleet
           sleet push ./pkgs
           sleet retention prune --package SharpHoundCommon --package SharpHoundRPC `
-            --release-labels 1 --dry-run
+            --disable --prerelease 5 --release-labels 1 --dry-run
--- a/.github/workflows/publish-dev-package.yml
+++ b/.github/workflows/publish-dev-package.yml
@@ -1,4 +1,3 @@
-name: Publish
 name: Publish Dev Package
 
 on:
--- a/.github/workflows/publish-dev-package.yml
+++ b/.github/workflows/publish-dev-package.yml
@@ -35,14 +35,50 @@ jobs:
           mkdir pkgs
           dotnet pack --no-restore -c Release -p:PackageVersion=${{ steps.version.outputs.dev_version }} -o ./pkgs
 
-      #the prune command deletes older package versions to avoid -dev package clutter
-      #it groups packages by -label, and keeps the 5 latest of each pre-release label (-rc1, -rc2, -dev, etc)
       - name: Publish to SpecterOps Packages
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_KEY }}
         run: |
           dotnet tool install -g sleet
-          sleet push ./pkgs
-          sleet retention prune --package SharpHoundCommon --package SharpHoundRPC `
-            --stable -1 --prerelease 5 --release-labels 1 --dry-run
+
+      #the prune command deletes older -dev package versions to avoid clutter
+      #it deletes any versions of the $packageIds with "-dev" that are older than the first $maxDevVersions
+      - name: Prune old -dev packages
+        shell: pwsh
+        run: |
+          $dryRun = $true
+          $packageIndexUrl = 'https://s3.amazonaws.com/bloodhound-ad/sleet.packageindex.json'
+          $packageIds = @('SharpHoundCommons', 'SharpHoundRPC')
+          $maxDevVersions = 2
+          
+          $packageIndex = Invoke-RestMethod -Uri $packageIndexUrl
+          
+          foreach ($packageId in $packageIds) {
+            $devPackages = @(
+              foreach ($version in ($packageIndex.packages.$packageId | Where-Object { $_ -like '*-dev.*' })) {
+                try {
+                  [pscustomobject]@{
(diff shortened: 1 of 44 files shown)


.github/workflows/publish-dev-package.yml
```diff
@@ -0,0 +1,40 @@
...
+      #it groups packages by -label, and keeps the 5 latest of each pre-release label (-rc1, -rc2, -dev, etc)
+      - name: Publish to SpecterOps Packages
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_KEY }}
+        run: |
+          dotnet tool install -g sleet
+          sleet push ./pkgs
+          sleet retention prune --package SharpHoundCommon --package SharpHoundRPC `
+               --stable 999 --prerelease 5 --release-labels 1
```
mykeelium: question: Should we prune stable at all? I understand 999 is a lot, but looking at the docs it seems like if it is not specified, it will follow the feed's settings. Is there a benefit to prune stable?

.github/workflows/publish-dev-package.yml
```diff
@@ -0,0 +1,40 @@
...
+          dotnet-version: 8.0.x
+
+      - name: Restore dependencies
+        run: dotnet restore
+
+      - name: Pack
+        run: |
+          $DATE_STAMP=$(date +%Y%m%d%H%M)
+          mkdir pkgs
+          dotnet pack --no-restore -c Release -p:PackageVersion=0.0.0-dev.$DATE_STAMP -o ./pkgs
```
mykeelium: question (non-blocking): Is it possible to add the version rather than the 0.0.0? or does this add unneeded complexity in the build and human processes?

.github/workflows/build.yml
```diff
@@ -1,25 +0,0 @@
...
-    runs-on: windows-latest
-    steps:
-    - uses: actions/checkout@v3
-
-    - name: Setup .NET
-      uses: actions/setup-dotnet@v3
-      with:
-        dotnet-version: |
-          7.0.x
-          5.0.x
```
mykeelium, line 16: note (non-blocking): I sincerely hope this is not needed anymore because I know it is way out of date, but I remember something about it being required to be built on this version. I may just be misremembering the readme. I will look into it.

mykeelium (changes requested):
Minor question on the stable retention, curious on the reasoning, otherwise it looks good to me!

.github/workflows/publish-dev-package.yml
lrfalslev: I'm not sure how the feed settings are configured, but when I try not specifying how many stable versions to retain sleet complains that it's a required argument.

image

.github/workflows/publish-dev-package.yml
lrfalslev: yeah, I think that will be a better approach. I'll do some tests publishing and consuming [version]-dev.$date_stamp. 

We'll want BED-3959 merged beforehand so we can pull the shared version for the projects with `dotnet msbuild Directory.Build.props --getProperty:Version`.

.github/workflows/build.yml
lrfalslev: I've been able to pull and use the libraries built with .net 8 fine so far. Though I've only ran test suites and haven't tried actually running collectors with the -dev libraries. 

I've previously built local changes to SHCommon on .net8 when working on stories and haven't had issues. I'll see if I can find anything that would restrict the sdk version for building.

.github/workflows/publish-dev-package.yml
mykeelium: it looks like you might be able to disable the stable pruning, I am a bit unsure why it is so arbitrary and difficult to do. Maybe my thinking on it is flawed. it seems like you would need to use `--disable` 

https://github.com/emgarten/Sleet/blob/main/doc/commands.md#options-10

lrfalslev: after more research I've come back to just setting a large stable retention number. If we're unhappy with that approach we could try something else but I don't think sleet offers the fine grained control for more specific pruning.

`sleet retention prune` requires both stable and pre-release to be > 0.
`sleet retention settings --disable` will simply clear/reset retention settings in sleet.settings.json where default values for retention commands are set

Potential paths forward:
- Writing our own pruning script is possible, but it looks like it'd be a little gnarly and would be harder to maintain.
- We can have a separate feed for -dev packages. That way we isolate pruning to only -dev and won't ever touch -rc or stable packages. But we'd have to setup and maintain a separate feed.
- We can run prune on the shared feed and just set a high number for stable, while regularly pruning -rc and -dev.

@mykeelium thoughts on the best approach?

mykeelium: I think your way of a large stable retention is the best approach. I am interested in another feed, I really do not believe stable should be pruned, but I will do more reading to see why it is so necessary. That being said, a high enough number should be fine for us to weigh the pros and cons of another feed. The pruning script seems to be more trouble than it is worth. Thank you for looking into this more.

lrfalslev: I was able to get a pwsh script working that doesn't feel too heavy. Pro is we won't touch stable or -rc versions. 
It parses the sleet.packageindex.json which sleet uses for feed management, and builds a list of -dev packages to target for deletion. 
I can revert back to sleet prune if we decide that's the best approach for now, while we consider separate feeds.

mykeelium (approved):

PR #297, opened 2026-04-15
chore: SHC -dev package workflows gracefully handle -rc packages BED-7878

lrfalslev:
## Description
Rewrite SHC -dev publishing workflow and SH/SHE csprojs to consume -dev packages, while gracefully handling -rc package versions.
Upgrade .net and testing package versions.
Update readmes.

## Motivation and Context

This PR addresses: [BED-7878](https://specterops.atlassian.net/browse/BED-7878)

## How Has This Been Tested?

See `Manual Testing Instructions` on ticket.

## Types of changes

- [x] Chore (a change that does not modify the application functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have met the contributing prerequisites
  - Assigned myself to this PR
  - Added the appropriate labels
  - Associated an issue: https://github.com/SpecterOps/BloodHound/issues/672
  - Read the Contributing guide: https://github.com/SpecterOps/BloodHound/wiki/Contributing
- [x] I have ensured that related documentation is up-to-date
  - Open API docs
  - Code comments
- [x] I have followed proper test practices
  - Added/updated tests to cover my changes
  - All new and existing tests passed

## Summary by CodeRabbit

* **Chores**
  * CI/CD workflows now use .NET SDK 10.0.x for building, testing, and publishing
  * Test projects retargeted to .NET 10.0
  * Upgraded test and coverage tooling to newer versions

* **Tests**
  * Removed deprecated smoke tests
  * Adjusted several tests to change timing/latency behavior and related assertions

[BED-7878]: https://specterops.atlassian.net/browse/BED-7878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/unit/AdaptiveTimeoutTest.cs
```diff
@@ -29,7 +29,7 @@ public async Task AdaptiveTimeout_GetAdaptiveTimeout_NotEnoughSamplesAsync() {
         var adaptiveTimeoutResult = adaptiveTimeout.GetAdaptiveTimeout();
         Assert.Equal(maxTimeout, adaptiveTimeoutResult);
         
-        Assert.InRange(observedLatency, 0.0, 100);
+        Assert.True(observedLatency >= 0.0);
```
lrfalslev: AdaptiveTimeout and Timeout tests were flaky in GH runners.
If observedLatency is positive callback was invoked, even if stressed runner took longer than timeout to evaluate.

test/unit/AdaptiveTimeoutTest.cs
```diff
@@ -79,15 +79,15 @@ public async Task AdaptiveTimeout_GetAdaptiveTimeout() {
         var minTimeout = TimeSpan.Zero;
         var adaptiveTimeout = new AdaptiveTimeout(maxTimeout, minTimeout, new TestLogger(_testOutputHelper, Microsoft.Extensions.Logging.LogLevel.Trace), 10, 1000, 3);
 
-        await adaptiveTimeout.ExecuteWithTimeout(async (_) => await Task.Delay(40), latencyObservation: LatencyObservation1);
-        await adaptiveTimeout.ExecuteWithTimeout(async (_) => await Task.Delay(50), latencyObservation: LatencyObservation2);
-        await adaptiveTimeout.ExecuteWithTimeout(async (_) => await Task.Delay(60), latencyObservation: LatencyObservation3);
+        await adaptiveTimeout.ExecuteWithTimeout((_) => Task.CompletedTask, latencyObservation: LatencyObservation1);
```
lrfalslev: On busy runner tasks could take longer than max timeout to run so adaptive timeout would stay at max and `Assert.True(adaptiveTimeoutResult < maxTimeout);` would fail

test/unit/TimeoutTests.cs
```diff
@@ -34,12 +34,12 @@ public async Task ExecuteWithTimeout_Timeout_Cancel() {
         var shouldRemainFalse = false;
         var timeout = TimeSpan.FromMilliseconds(100);
         var func = (CancellationToken t) => {
-            Thread.Sleep(TimeSpan.FromMilliseconds(500));
+            Thread.Sleep(TimeSpan.FromMilliseconds(1000));
```
lrfalslev, line 37: Increased timeout to ensure that busy runner has time to execute the timeout's cancel function before checking if cancelled.

test/unit/AdaptiveTimeoutTest.cs
mykeelium: @definitelynotagoblin do you have any notes on this? I know this is a problem area for sure, but it makes me wonder if this test is even doing anything if we move to this paradigm. Maybe that is due to the nature of the function, but I am not sure what would be the best way to handle this. I almost lean to a more forgiving timeout.

RPCTest/RPCTest.csproj
```diff
@@ -1,18 +1,19 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>net8.0</TargetFramework>
+        <TargetFramework>net10.0</TargetFramework>
```
mykeelium, line 4: It might be valuable to consider multitarget here to include net472. It will definitely cause some issues, but currently our tests are testing functionality that does not actually represent our built library. For instance, one change in this pr, the X509CertificateLoader.LoadCertificate is a .NET 9+ and would not be representative of the code path that actually runs in the shipped library when targeting net472. When the library runs, it is still using new X509Certificate2(byte[]) so the mock would not catch any issues in that functionality directly. In addition there are many functionality differences in System.DirectoryServices, which may be missed in our testing. 

This might open a larger discussion since this is not really the scope of the this PR, but it may be helpful to be done sooner rather than later to cut down on the pain in support.  

refs:
our current functionality is obsolete: https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0057
In order to get the loader class in 472, we would have to have the package: https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificateloader?view=net-10.0#applies-to

mykeelium (commented):
I am cool with this moving forward, I think we should at least decide on the notes, but they shouldn't be blocking to this. Let me know if/when I will mark approved, I got burned with auto merge.

RPCTest/RPCTest.csproj
lrfalslev: For now reverted the change in MockDirectoryObject, to use the old X509Certificate2 constructor, as it's just a warning in .net9+. I'm open to revisiting these tests later to investigate discrepancies. I am curious with all the facades we have how much the tests are actually exercising the SUT

test/unit/AdaptiveTimeoutTest.cs
lrfalslev: Current version uses Thread.Sleep in place of Task.Delay so we don't yield the thread and avoid much larger than expected delays, while still simulating async operations.

mykeelium (approved):

